PR #85, opened 2014-12-02
Adding global test fail/ignore message, using safe sprintf alternatives, adding configurable test result token delimiter minor clean up

ProudGecko:
1. Added UnitySetGlobalMessage(), TEST_SET_GLOBAL_MESSAGE(), TEST_RESET_GLOBAL_MESSAGE(), and const char* UnityStrGlobalTestMsg to allow a test (mostly useful for lookup table style tests) to specify a global message to be printed for any TEST_ASSERT/TEST_FAIL/TEST_IGNORE/CMock failure.  Implementation of this includes refactoring of UnityAddMsgIfSpecified() and addition of UnityAddLonelyMsgIfSpecified() to handle assert specified messages and the global message.
2. Added call to TEST_RESET_GLOBAL_MESSAGE() in create_runtest() to clear a previously set global test message.
3. Added UNITY_RESULT_DELIMITER #define which allows for delimiters other than ':'.  This was needed by our CI build because the Unity results were being detected as GCC warnings.
4. Added SAFE_SPRINTF() macro to abstract snprintf() vs sprintf_s() usage instead of the less safe sprintf()
5. Minor clean up for readability/defensive coding

### DIFF
--- a/auto/generate_test_runner.rb
+++ b/auto/generate_test_runner.rb
@@ -195,7 +195,7 @@ class UnityTestRunnerGenerator
       end
       output.puts("}\n")
 
-      output.puts("void CMock_Verify(void)")
+      output.puts("static void CMock_Verify(void)")
       output.puts("{")
       mocks.each do |mock|
         mock_clean = mock.gsub(/(?:-|\s+)/, "_")
--- a/auto/generate_test_runner.rb
+++ b/auto/generate_test_runner.rb
@@ -195,7 +195,7 @@ class UnityTestRunnerGenerator
       end
       output.puts("}\n")
 
-      output.puts("static void CMock_Verify(void)")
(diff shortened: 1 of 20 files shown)


mgiaco:
Overall it would be great to use snprintf but it is not available on every tool chain. On Keil C166 as an example exists no implementation for snprintf. I could write my own of course. So I think there should also be a switch for save or not save.

mvandervoord:
I hate to do this, but there are too many unrelated changes in this pull request. Some of them are failing our regression tests and some of them seem to be working. I'm going to have to reject this pull request at this time. I would really love it if you were to resubmit it in chunks... for example, the test delimiter option is a nice simple and useful feature. The global message feature is an interesting one that I think people would appreciate too!

PR #95, opened 2015-01-17
Make Unity's compilation flags stricter

trianglee:
Unity is used in multiple environments, with different compilers and different default warning levels.
This change makes the default compiler switches used in tests and examples (for gcc, at least) more strict - and adapts the code to cleanly compile with the extra warnings.
No behavior changes.


jwalkerbg:
I would like to propose putting prototypes of the functions made static in unity.c:

static void UnityAddMsgIfSpecified(const char* msg);
static void UnityAddMsgIfSpecified(const char* msg)
{
.....
}

MISRA rules, par example, require prototypes to be seen before function body definition.

trianglee:
I personally find this MISRA rule quite weird, for static functions (it increases chance of mistakes, instead of decreasing them).
But if MISRA compliance is needed, this is something that can be added. Does the code otherwise pass MISRA compliance?

mvandervoord:
Feel free to submit further changes for the MISRA rules... but lets keep them as separate requests. It's best to keep these pull requests as slim as possible.

PR #96, opened 2015-01-20
Added "const" to one argv parameter that was missed when adding strict error checks.

trianglee:
One argv was missing a "const". It caused "example_2" to not compile properly.


trianglee:
Any update with this pull request? It is a very simple one, and fixes a broken example.

PR #103, opened 2015-02-16
Added assertions counter.

jwalkerbg60:
We have global test counter and this is OK - we know at the end how many tests have passed and failed. However, tests may contain more than one TEST_ASSERT macro calls. It could be interesting for one to know the total number of assertions passed.


mvandervoord:
I like the concept of this, but Unity is now tied into numerous systems which are relying on a particular output. If you rework this so that (a) it is optional and (b) all the tests pass, I'd love to merge it.

PR #108, opened 2015-04-23
New asserter: TEST_ASSERT_EQUAL_STRING_LEN

eivindt:
Adds support for asserting that strings are equal up-to n characters.

Adds the same functionality that strncmp() offers, asserting that the n first characters of the expected/actual strings are the same, ignoring the rest of the characters.

TEST_ASSERT_EQUAL_STRING_LEN and TEST_ASSERT_EQUAL_STRING_LEN_MESSAGE has been added.

Unit tests for TEST_ASSERT_EQUAL_STRING_LEN has been added following the same pattern as for TEST_ASSERT_EQUAL_STRING.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,9 +169,17 @@ String Assertions
 
 Compare two null-terminate strings.  Fail if any character is different or if the lengths are different.
 
+    TEST_ASSERT_EQUAL_STRING_LEN(expected, actual, len)
+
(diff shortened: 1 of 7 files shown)


mvandervoord:
Ah, what a good idea. I'll check this out and then merge it. :)

PR #109, opened 2015-05-22
Reverting of redundant castings, Fixed -Wcast-qual warnings

laborer2008:


mvandervoord:
Sorry, the void* cast is in there to keep some embedded compilers from erroring out on this cast. I realize that it's those embedded compilers that are in err, but keeping it this way allows everyone to always compile (with an occasional warning).

laborer2008:
What compilers are unable to compile such a code?

mvandervoord:
I just went spelunking through all the commits to see why they were added in the first place. I'm not entirely clear on the reasoning (they were submitted with some other changes that were the main point of the update). 

I'm going to therefore assume no one will mind if I merge this and we'll see who complains. ;)

Thanks

Mark

laborer2008:
Thanks,

I also thought to propose an alternative:

Put before UnityAssertEqualIntArray():
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wcast-qual"

And after it:
#pragma GCC diagnostic pop

It may be used if someone complaining about this changeset show up.
Any gcc incompatible compiler knows nothing about "pragma GCC" and therefore should just ignore it.

Issue #61, opened 2014-03-05
Wrong value reported when assert fails


rasup:
Example:

```
TEST_ASSERT_EQUAL_UINT8( 0xFFU, 0x01U );
```

Gives output:
_:FAIL: Expected 4294967295 Was 1_
I expected it to be:
_:FAIL: Expected 255 Was 1_

The solution is to use the correct display style in unity_internals.h:

```
#define UNITY_TEST_ASSERT_EQUAL_UINT8(expected, actual, line, message)                           UnityAssertEqualNumber((_U_SINT)(_US8 )(expected), (_U_SINT)(_US8 )(actual), (message), (UNITY_LINE_TYPE)line, UNITY_DISPLAY_STYLE_UINT8)
#define UNITY_TEST_ASSERT_EQUAL_UINT16(expected, actual, line, message)                          UnityAssertEqualNumber((_U_SINT)(_US16)(expected), (_U_SINT)(_US16)(actual), (message), (UNITY_LINE_TYPE)line, UNITY_DISPLAY_STYLE_UINT16)
#define UNITY_TEST_ASSERT_EQUAL_UINT32(expected, actual, line, message)                          UnityAssertEqualNumber((_U_SINT)(_US32)(expected), (_U_SINT)(_US32)(actual), (message), (UNITY_LINE_TYPE)line, UNITY_DISPLAY_STYLE_UINT32)
```

Instead of:

```
#define UNITY_TEST_ASSERT_EQUAL_UINT8(expected, actual, line, message)                           UnityAssertEqualNumber((_U_SINT)(_US8 )(expected), (_U_SINT)(_US8 )(actual), (message), (UNITY_LINE_TYPE)line, UNITY_DISPLAY_STYLE_UINT)
#define UNITY_TEST_ASSERT_EQUAL_UINT16(expected, actual, line, message)                          UnityAssertEqualNumber((_U_SINT)(_US16)(expected), (_U_SINT)(_US16)(actual), (message), (UNITY_LINE_TYPE)line, UNITY_DISPLAY_STYLE_UINT)
#define UNITY_TEST_ASSERT_EQUAL_UINT32(expected, actual, line, message)                          UnityAssertEqualNumber((_U_SINT)(_US32)(expected), (_U_SINT)(_US32)(actual), (message), (UNITY_LINE_TYPE)line, UNITY_DISPLAY_STYLE_UINT)

```

_Edit: Was a little to fast with the first solution, the one above should fix it_
[...]

PR #111, opened 2015-06-05
Fix to prevent warning using armcc: Warning:  #61-D

emidttun:
The armcc compiler gives a warning on the changed line:
Warning:  #61-D: integer operation result is out of range.
Really easy to fix.

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -135,7 +135,7 @@ void UnityPrintNumber(const _U_SINT number_to_print)
     _U_SINT next_divisor;
     _U_UINT number;
 
-    if (number_to_print == (1l << (UNITY_LONG_WIDTH-1)))
+    if (number_to_print == (1ul << (UNITY_LONG_WIDTH-1)))
     {
         //The largest representable negative number
         UNITY_OUTPUT_CHAR('-');


mvandervoord:
number_to_print is signed, so changing 1l to 1ul creates a comparison of signed and unsigned values.

The result being out of range is part of the point of that statement if you've reached the top of that number... maybe there is a cleaner way we can fix this though?

emidttun:
How about using LONG_MIN from limits.h? I found it defined in both armcc and gcc.

mvandervoord:
We've run into problems using things from limits.h before. I believe Keil and Microchip don't have a limits.h with their compilers. 

Still, you bring up a really good point. Is LONG_MIN a define? I COULD check if LONG_MIN exists... and if it doesn't, create one using the method I have here... then platforms that have LONG_MIN would always have the right value and you'd just be left with a possible minor warning for other platforms. 
What do you think?

emidttun:
This is how it is implemented in armcc:  
```
    /* maximum value for an object of type unsigned int */
#define LONG_MIN  (~0x7fffffffL)
```
(Btw, it is 'Keil', but for Cortex-M series. Might be different for C51 for example.)

It might be a better definition for comparison since the ```<<``` operator is avoided. Last point: the tool chain is for 32-bit; I think LONG_MIN is defined differently for 64-bit.

mvandervoord:
Yeah. Not knowing the int size of the toolchain is the reason we have such a generic derivation. Your example with the inversion is interesting though. I'll mull that over and see if I can come up with something cleaner based on that.

Thanks for the ideas.

PR #121, opened 2015-07-14
On assert, print the file name where the assert happened.

trianglee:
Previously, the file name where the test is defined was printed, which is
far less useful and confusing. Line number was always of the
correct file name.

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -328,13 +328,14 @@ static void UnityTestResultsBegin(const char* file, const UNITY_LINE_TYPE line)
 }
 
 //-----------------------------------------------
-static void UnityTestResultsFailBegin(const UNITY_LINE_TYPE line);
-static void UnityTestResultsFailBegin(const UNITY_LINE_TYPE line)
+static void UnityTestResultsFailBegin(const UNITY_LINE_TYPE line, const char* file);
+static void UnityTestResultsFailBegin(const UNITY_LINE_TYPE line, const char* file)
 {
 #ifndef UNITY_FIXTURES
-    UnityTestResultsBegin(Unity.TestFile, line);
+    UnityTestResultsBegin(file, line);
 #else
     UNITY_UNUSED(line);
+    UNITY_UNUSED(file);
 #endif
     UnityPrint(UnityStrFail);
     UNITY_OUTPUT_CHAR(':');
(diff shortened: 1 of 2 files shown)


mvandervoord:
Hi.

Thanks for the submission. I understand where this submission is coming from, unfortunately it doesn't do what we want it to. We originally started with a configuration similar to this, but have since switched away from it intentionally. 

Just as unit testing is supposed to encourage good design of code, we're trying to encourage good test design when people write tests with Unity. This includes keeping tests to-the-point and grouped in a single test file.

We're glad to have external helpers (and those of us who use CMock use a LOT of automatically generated "helpers" in the form of mocks. But in these cases we always want to know what line and test file failed, not where in the helper it failed. Your helpers should be well written enough to announce what exactly went wrong, so it should not be necessary to jump into the helper to look. Therefore the test line and file is what we've decided to show.

At this point you may be thinking that the test line number is wrong. In that case, I'd suggest checking out how the TEST_ASSERT_ macros wrap the UNITY_TEST_ASSERT_ macros. If you design helpers the same way (which CMock does), then the proper line number will always get passed along as needed.

I hope this explains clearly why things are the way they are. I'd be happy to discuss if further if something is still unclear (or even if you just disagree with me). Feel free to write me. :)

Again, THANK YOU for your work. I'm sorry to have to say no to this particular patch. Your work seemed thorough and high-quality.

Mark

trianglee:
Hi Mark,

Thanks for your elaborate reply.

Indeed, the use-case you are outlining is exactly what triggered us to add this kind of information.
An helper is running and fails (and uses Unity assertions). Without the file-name, all we are told is the line-number (and, it is the wrong line number for the file-name shown, because the helper is a function, not a macro).
(I'm not sure how we can get a correct line number in this case - I must be missing something).

One way to avoid that is to never call Unity assertions from helpers - make the helpers return a success/failure indication, and fail only in the main test body. This hides very useful information in case there is more than a single assertion in the helper, and in our case this is very common. We just end up adding a lot of non-Unity prints to the helpers, which defeats the purpose of the Unity macros.
 
Would you consider allowing for both options, making this behavior configurable? That is, by default it would print the file-name of the test, but a compilation flag would allow asking for the file-name of the failing file? (this would require a simple extra change in UnityTestResultsFailBegin, which should be straight forward).
Otherwise, we would be forced to maintain this as an independent patch which is a real shame - as it touches nearly every Unity function, and makes upgrading a real annoyance...

Thanks,

  -- Nimrod

PR #122, opened 2015-07-14
Added ability to get number of ignored tests and total number of tests.

trianglee:


trianglee:
We are running Unity in a few different setups, some of them make it very hard to mechanically parse the tests output.
It is possible to get the logs for a human to browse if things go wrong, but this is a fairly painful process, and the resulting log is often hard to parse due to "noise" - the logging mechanism adds timestamps and other indicators that we can't disable.
To overcome this, we wrap the call to Unity with some code that sends out the vital statistics (successful tests, failed tests, ignored tests) to the outside world.

mvandervoord:
Makes sense to me. Thanks for elaborating!  :)

PR #131, opened 2015-09-09
Change comments style to compile with std=c89

andyhelp:


andyhelp:
Hi,
Is it something that could be merged in?
Thanks.

mvandervoord:
Hi Andy.

Out of curiosity, what compiler are you using that requires this level of backwards compatibleness? We've encountered a lot of quirky compilers with this project, but I've not yet seen one that requires old-school comments.

andyhelp:
Hi Mark,

The issue is with integrated embedded SDK + IDE.
IDE runs gcc with ```-ansi``` flag. 

There seems to be no option to disable it, although I would prefer to keep the flag enabled in both production code and utests with unity.

Thanks.

bjones1:
Personally, I find C89 comments less readable. I think that supporting C89-only compilers is a corner case and wouldn't encourage making this a part of the code base. In this specific case, a makefile without the -ansi flag would seem to provide an alternative to running the compiler from the IDE.

andyhelp:
In my case IDE is the only supported way of building the project and IDE autogenerates makefiles. Maintaining separate makefile will be difficult, e.g. to keep it in sync with original one.

Good part of the patch is that it makes Unity applicable to more projects/users and is backward compatible.

benwainwright:
Is this likely to be merged in? I really need c90/c89 compatibility for the university course I'm doing

xor-gate:
Personally for higher portability supporting c89 style comments is good for people who use it with exotic or custom build compilers which have no real support for single line (c99 style) comments. As the project doesn't mention any codestyle it is both correct. When looking at a widely implemented codestyle e.g Linux codestyle we should stick with C89 `/* */`  comments. Even better would be to enforce pull requests and changes to the codebase by using a linting tool, e.g checkpatch (when enforcing linux codestyle). I'm not against another codestyle but for the linux codestyle there is a good document plus a tool to enforce it.

wolf99:
This would be great as the MISRA-C standard also requires `/* .. */` style comments over `//..` - safety being probably a higher concern than readability in most embedded systems.

mvandervoord:
Actually, MISRA-C 2012 updated to support // specifically BECAUSE they learned that it made code more readable and therefore more maintainable.

wolf99:
@mvandervoord Ah, I was unaware of this development! Guess older compilers or `-ansi` would still require it though :\

mvandervoord:
:) Yeah. I have a strong internal struggle at the moment between my desire to keep things clean and readable and my desire to support as many compilers out there as possible. I've even considered including a script to convert to /* */ for those who want that, but continue to deliver as //... but that might be taking things a little too far. ;)

xor-gate:
@mvandervoord keep in mind with research/custom or damn old compilers maybe only C89 is supported.

wolf99:
@mvandervoord  I do find that once you have an editor that does even halfway decent syntax highlighting for comments and all comments are in one style or the other and other things stylistic choices such as indents, block comment fencing, etc, are adhered to consistently, then either style is very readable.
For instance `//` comments are more likely to be used on the same line as code, but if adhering to a specific line limit then this is less likely as lines become to long. i.e.:

    /* My awesome comment. */
    MyAwesome.Line = OfCode;

is little different to:

    // My Awesome comment.
    MyAwesome.Line = OfCode;

TLDR: if consistent style is adhered to either style can be very readable.

xor-gate:
There is also no codestyle linting for Unity currently. And there is no codestyle guideline. This also covers comment style. E.g linux `checkpatch` & codestyle guidelines, it needs some tweaking for deep embedded software but there is at least a tool to check consistency. IMHO because Unity can be used in deep embedded systems we should support `ansi`-only compilers and get rid of single line C99 comments.

mvandervoord:
;) alright, guys. You have convinced me of two things:
(1) we should switch to /* .. */ comment style
(2) we should formalize the ThrowTheSwitch.org coding style. 

To address #1, we need to update this branch to pull in the latest changes (it currently has conflicts). Any volunteers to clean up this branch so we can merge it properly?

To address #2, I'll start putting something together ASAP. I'll put together a guideline as well as a verification of that style

xor-gate:
@mvandervoord probably you or I should open a new issue to resolve (2). Then we can discus the coding style enforcement during CI phase with a lint tool (checkpatch, cpplint etc). Else this pull request is abused for discussion :+1:

mvandervoord:
done. (see issue #194)

andyhelp:
Is used comment style ok? If yes, then I can rebase.

mvandervoord:
Yes, the proposed style is good. Even if it ends up getting tweaked over time, it's a good starting point and it fulfills the requirement for those who need it now.

andyhelp:
done. 
I checked that example_1 builds fine (now with std=c89).

example_1 report FAIL, but it fails in the same way before the change.
```
test/TestProductionCode.c:61:test_FunctionWhichReturnsLocalVariable_ShouldReturnCurrentCounter_ButFailsBecauseThisTestIsActuallyFlawed:FAIL: Expected 0x00001234 Was 0x00005A5A

-----------------------
5 Tests 2 Failures 0 Ignored 
FAIL
make: *** [default] Error 2
```

PR #140, opened 2015-10-22
add newline before filename on error

cosmicexplorer:
many editors such as emacs use regular expressions to parse
filenames. without adding a newline, emacs thinks the file the error
occurs at is named ".file" instead of "file", because unity outputs
a dot on every successful test run. This makes it difficult to jump to the error.

Old output:
```
.test/siman-test.c:68:TEST(siman, EqualsGSL):FAIL: Values Not Within Delta . should be 1.363131
```

Emacs reads this as a file named ".test/siman-test.c".

New output:
```
.
test/siman-test.c:68:TEST(siman, EqualsGSL):FAIL: Values Not Within Delta . should be 1.363131
```

Emacs reads this as the correct filename.

This won't leave ugly newlines in between failed assertions, because Unity
immediately quits the test when an assertion fails.

New output (multiple failures):
```
.
test/siman-test.c:68:TEST(siman, EqualsGSL):FAIL: Values Not Within Delta . should be 1.363131
.
test/siman-test.c:68:TEST(siman, EqualsGSL):FAIL: Values Not Within Delta . should be 1.363131
```


cosmicexplorer:
Ok, here's an attempt at making the error messages more useful.
```
# failure (old --- new)
.test/siman-test.c:68:TEST(siman, EqualsGSL):FAIL: Values Not Within Delta . should be 1.363131
---
test/siman-test.c:68:TEST(siman, EqualsGSL):FAIL: Values Not Within Delta . should be 1.363131
# non-verbose success (old --- new)
.
---

# verbose success (old --- new)
TEST(siman, EqualsGSL) PASS
---
TEST(siman, EqualsGSL):PASS
```

cosmicexplorer:
Sorry for the reversions. I can't figure out how to maintain the single dot for non-verbose output; because `fixtures` doesn't know if the next test will have an error or not, it simply always has to output a newline, just in case the next test could output an error. This means we can't have the stream of dots `....` on the same line for non-verbose output and also not mangle the file names with leading dots (as shown in the original post). The alternatives I can see are 1. having no dots at all or 2. having a newline in between each dot. The dots can be helpful for showing that the tests aren't hanging though, so a compromise could be showing a single dot per test group, perhaps? This would make non-verbose output smaller but still show progress is occurring. Currently, it just outputs no dots, but that change is easy to make.

cosmicexplorer:
Hey, I just added a commit that shows what adding a period and newline on non-verbose success would look like. Would you be interested in merging any form of these changes? My project just has my fork as a submodule so time is not an issue, I'm just interested in knowing if there's a chance of merging this.

Output on success, success, failure in a single test group looks like:
```
.
.
test/siman-test.c:80:TEST(siman, NotEqualsGSL):FAIL: Values Not Within Delta . should be 1.363131
```

PR #141, opened 2015-10-28
Fix some C++ compatibility issues

adiorion:
I'm using Unity in a project with a mix of C and C++ tests, and these two issues tripped me up.
 - without a `extern "C"` prototype, `setUp()` ends up being mangled in my test_foo.cpp build and thus unity.c invokes its local weak symbol instead.
 - without wrapping unity.h in `extern "C"`, functions like `UnityAssertEqualNumber` get mangled and the link fails with `undefined reference to `UnityBegin(char const*)'`.

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -1231,9 +1231,9 @@ void UnityIgnore(const char* msg, const UNITY_LINE_TYPE line)
     UNITY_WEAK_ATTRIBUTE void tearDown(void) { }
 #elif defined(UNITY_WEAK_PRAGMA)
 #   pragma weak setUp
-    void setUp(void);
+    void setUp(void) { }
 #   pragma weak tearDown
-    void tearDown(void);
+    void tearDown(void) { }
 #else
     void setUp(void);
     void tearDown(void);
--- a/src/unity.h
+++ b/src/unity.h
@@ -15,6 +15,9 @@ extern "C"
 
 #include "unity_internals.h"
 
+void setUp(void);
+void tearDown(void);
+
 //-------------------------------------------------------
 // Configuration Options
 //-------------------------------------------------------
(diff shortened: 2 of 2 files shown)


mvandervoord:
The first set of changes (with the extern C) looks great. I honestly thought I had added that ages ago (but clearly did not... or it got lost in the shuffle at some point).

The second set of changes might require a little thought. Have you verified it works with all three variants of weak definition in Unity.c around lines 1227-1240?

adiorion:
Great question! One question back -- I don't see how UNITY_WEAK_PRAGMA works, it seems like it declares `setUp` to be weak but never defines it, which I think means there's no symbol `setUp` in the final unity.o (neither weak nor strong). At least, that's what some quick testing with x86_64 gcc 5.2.1 seems to show.

I think UNITY_WEAK_PRAGMA needs function bodies, like WEAK_ATTRIBUTE has:
```
@@ -1232,9 +1232,9 @@ void UnityIgnore(const char* msg, const UNITY_LINE_TYPE line)
     UNITY_WEAK_ATTRIBUTE void tearDown(void) { }
 #elif defined(UNITY_WEAK_PRAGMA)
 #   pragma weak setUp
-    void setUp(void);
+    void setUp(void) { }
 #   pragma weak tearDown
-    void tearDown(void);
+    void tearDown(void) { }
 #else
     void setUp(void);
     void tearDown(void);
```
but, maybe I'm missing something about other C compilers (Solaris and AIX would be good to test this with). Unfortunately I don't have anything but gcc and clang handy.

adiorion:
here's a proposed fix for the issue with UNITY_WEAK_PRAGMA, if my analysis is correct. LMK if I'm missing anything (and if you have any way to test on vendor compilers!).

mvandervoord:
:) Makes sense to me. I'll have to check out my test rig and see why I wasn't catching that problem. I really thought I had both of those cases covered already, but maybe I just had the first? In any case, thanks for the catch!

PR #156, opened 2016-01-20
Two small changes to runner generation

pmendham:
These two commits are two small changes we have made locally to allow us to use Unity.
1. Added a command line option to allow us to rename main. This is to allow integration with other system and/or use on a platform with a different entry point. Only the name is changed (not the type signature) for simplicity.
2. Improved the handling of generated header files to make them create valid prototypes, and function as standalone headers (include dependencies on the framework). Also I cleaned up the guard symbol as we have files in deep subdirectories and it was unwieldy.


pmendham:
Thank you!

PR #160, opened 2016-01-28
Added support for mocks in sub-directories.

pmendham:
This supports mock headers of the form:
    #include "some/dir/MockMyCode.h"
Where the mock name is actually MockMyCode.

I *think* this is the most common scenario when working with mocks
in subdirectories but if not this could be modified to support
alternate schemes. This works with our use of Unity and CMock and
our code base which makes heavy use of sub-directories.

### DIFF
--- a/auto/generate_test_runner.rb
+++ b/auto/generate_test_runner.rb
@@ -146,8 +146,9 @@ class UnityTestRunnerGenerator
 
   def find_mocks(includes)
     mock_headers = []
-    includes.each do |include_file|
-      mock_headers << File.basename(include_file) if (include_file =~ /^mock/i)
+    includes.each do |include_path|
+      include_file = File.basename(include_path)
+      mock_headers << include_path if (include_file =~ /^mock/i)
     end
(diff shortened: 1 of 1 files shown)


pmendham:
Dear ThrowTheSwitch,

Is this something that could be merged in? I don't believe it will affect others, as the current test runner generation completely strips sub-directories from include paths so it shouldn't break backwards compatibility unless that behaviour is desired in some way. At the moment we have a local patch for this which we use together with your latest release as I would rather not maintain a fork. I am happy to make changes if I understand the requirements,

Many thanks,
Peter

mvandervoord:
Hi.

Thanks for the helpful patch!  I've added one comment on a way to clean it up a little bit. If you implement that, I'd happily merge it in.  :)

Thanks again for your help!

Mark

pmendham:
Thanks for the feedback @mvandervoord I've made the change you suggested (I believe). If I've misunderstood or there is anything else you'd like me to change/improve please let me know.

Thanks!

Peter

PR #167, opened 2016-02-18
UnityPrintFloat buffer overflow fix

pacheco017:
`UnityPrintFloat()` overflows on some inputs, eg. `FLT_MAX`. Fix this by using `strncpy` instead of `strcpy`.

Additionally, `UnityPrintFloat()`'s temporal buffer is now large enough to allow formatting any single precision float without truncation (truncation still performed for certain double precision floats).

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -283,7 +283,7 @@ void UnityPrintMask(const _U_UINT mask, const _U_UINT number)
 void UnityPrintFloat(_UF number)
 {
     char TempBuffer[32];
-    sprintf(TempBuffer, "%.6f", number);
+    snprintf(TempBuffer, sizeof(TempBuffer), "%.6f", number);
     UnityPrint(TempBuffer);
 }
 #endif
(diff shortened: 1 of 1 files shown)


mvandervoord:
Hi.

Thanks for the helpful submission. It has a couple of minor problems. If you have a moment, would you mind sorting them out and then asking me to pull it again?

(1) We can't rely on float.h (and therefore the macro you specified) existing. We work with a lot of embedded compilers and (even when floats are supported) that file doesn't necessarily exist, sadly.
(2) There's no sense in making that buffer larger for non-floating point platforms (who also care a lot about wasted space).

My proposal:

We add another define like UNITY_NUMBER_BUFFER_SIZE or something like that. We allow people to set it themselves. If they don't we drop in a logical default: 32 if no floating point or something sufficiently large for float/double support. 

What do you think?

pacheco017:
No problem at all!

 I've opened #172 which is based on your suggestions. Please tell me if that's what you meant.

For future reference: would you have prefered me to keep addding commits to this branch, amended some commits, opened a new PR (as I've done), or something else?

Thanks for your time.

mvandervoord:
Thanks. I thought your fix in #172 was perfect. Just what I was thinking!

In the future, I think it's easiest if you just continue editing the same branch when it's a minor change like the one we were discussing... but I don't really feel strongly about that. This was easy too.

Thanks for the help! I really appreciate it!

pacheco017:
Thank you!

PR #177, opened 2016-02-29
Fix unreachable code in Unity on fully covered switch statements

jsalling:
The clang compiler warns here with `-Wunreachable-code`
The enum's switch statement covers all cases, so default is unused
Leave the break in the code as a comment, to be more clear
The history is this default case was added in f6bb7162 - compiler warning.
Then the break was added in c6dc96f3.


mjago:
Another option would be to add `UNITY_FLOAT_INVALID` to `UNITY_FLOAT_TRAIT_T`  

This enables default to catch invalid `style` values and bail. Whilst this is mainly for the compilers benefit it also gives the test system a failure to test against.

mvandervoord:
Yeah, I think we should probably go with something more like Martyn's suggestion. The default conditions are on there in the first place because a compiler was complaining about not having them there.  ;)

jsalling:
I like this idea, too. I did some reading; let me point out what I found. Overall, my suggestion is to merge this PR, and then decide how we should approach `enums` and `defaults` as standard practice. In either case, I'll even volunteer to work on a follow up commit to add or revert this one.

First, to clarify, this patch didn't get rid of the default case, it just deletes unused code flagged by my clang version as `unreachable-code`. There are really two issues here: compiling using `clang` `-W unreachable-code`, and also how to manage covered `default` cases in Unity in general.

In addition to the benefits mentioned, I think an `INVALID` would get rid of `covered-switch-default` warnings in clang. Right now, though, if you add to the `enum` and run `UnityAssertFloatSpecial(1.0f, NULL, __LINE__, UNITY_FLOAT_INVALID);` Unity _crashes_. **See footnote.

However, on the other side of the debate is whether to have `enum` `default` cases at all. You can't get rid of all the possible warnings, since `switch` and `switch-enum` have complimentary meanings in `gcc`. The code did work without a `default` case, but someone probably compiled with `-Wswitch-default` and that is how Unity is compiling now, by default. (Joke: How many defaults does it take...)

Quote from the [GCC manual](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wswitch_002denum-303):
>**-Wswitch**
>Warn whenever a switch statement has an index of enumerated type and lacks a case for one or more of the named codes of that enumeration. (The presence of a default label prevents this warning.) case labels outside the enumeration range also provoke warnings when this option is used (even if there is a default label). This warning is enabled by -Wall. 
>**-Wswitch-default**
>Warn whenever a switch statement does not have a default case. 
>**-Wswitch-enum**
>Warn whenever a switch statement has an index of enumerated type and lacks a case for one or more of the named codes of that enumeration. case labels outside the enumeration range also provoke warnings when this option is used. The only difference between -Wswitch and this option is that this option gives a warning about an omitted enumeration code even if there is a default label. 

See also the [LLVM Coding Standard for enums on covered switches](http://llvm.org/docs/CodingStandards.html#don-t-use-default-labels-in-fully-covered-switches-over-enumerations) and this stackexchange about [clang warnings with covered switches](http://programmers.stackexchange.com/questions/179269/why-does-clang-llvm-warn-me-about-using-default-in-a-switch-statement-where-all). Best of all, the 4th comment on this [GCC bug report](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61864) shows how clang & GCC differ and when each warning happens.

** Footnote:
If you are following along with the code, it crashes at 787:`UnityPrint(trait_names[trait_index]);` after passing `string=0x4 <Address 0x4 out of bounds>`. ...on my system, using `gcc`. I can fix this of course, but it's not as simple as adding one line to the `UNITY_FLOAT_TRAIT_T` enum.

jsalling:
I do like the idea, so I've pushed another commit using it.

My take-away from the post above is to build with `-Wswitch` (enabled by `-Wall`) and `-Wswitch-default`, but *not* the alternate `-Wswitch-enum`. It's how the rakefile build is done today. This mandates the use of `default` cases, but does not require all enumerated named codes to have a `case` label.

Adding an `enum` code to `UNITY_FLOAT_TRAIT_T` means the compiler will warn with `-Wswitch-enum`, however it also nicely fixes the unreachable code condition I was originally trying to solve, under `clang`.

mjago:
Apologies for not responding earlier - I've been rather busy. Your solution and test look great to me.

mvandervoord:
I agree too. And thanks for redoing that. I like this solution much better.

jsalling:
Thank you both. I appreciate the ideas and the reviews.

PR #180, opened 2016-04-11
Add command line arguments support

shellyniz:
## Added the following test runner generation options:

**auto_main_name**            - Generates the main method's name according to the test file name (bla_bla_test.c -> bla_bla_main). false by default.
**main_with_params**          - Allows command line arguments (argc, argv). false by default.
**main_export_decl**            - Allows export declaration to main (when using visual studio, "__declspec( dllexport )" may be useful). Empty by default.
**isolation_directive**            - Generates a precomplier directive to isolate the tested unit, based on the test file name  (ifdef _USE_REAL_TESTED_UNIT_NAME_). false by default.

### The command line arguments support handles the following flags:
**-l**                 : Lists all test methods. No test methods are executed.
**-f=_filter_**       : Executes only test methods which match the filter. 
                     The filter may be of the following formats:
                    test_case_name.*                             - all test methods in this test case
                    test_case_name.test_method_name  - only the selected test (including test method input parameters, if any)
                     empty filter                                           - all tests              
                     Filters can be cascaded, delimited by "\n\r".

### Added the following methods:
IsFlagTrue  - Analyzes flags passed in command line arguments (argc, argv).
MatchFilter - Determines if current test method name matched the filter passed in cmd line arguments.

### Added the following flags defeinitions
UNITY_FLAG_LIST_TESTS ('l')
UNITY_FLAG_FILTER_TESTS ('f')


mvandervoord:
This feature was merged (in a slightly different format) as f276c79

PR #185, opened 2016-04-25
Add Unity Fixture to the travisCI build and restore header declaration of UNITY_OUTPUT_CHAR

jsalling:
* Prevent changes in core Unity from silently breaking Fixture
* Restore default behavior of outputting a declaration of `UNITY_OUTPUT_CHAR()` in the internal header.
* Adding an option to remove the declaration, as non-default behavior - use `UNITY_OMIT_OUTPUT_CHAR_HEADER_DECLARATION`

Note: The first commit (adding Fixture to travisCI) will fail the build, as the Fixture build is currently broken. The next push will fix this up.

Discussion:
The internals header was declaring UNITY_OUTPUT_CHAR if it was overridden by the user. This was changed in PR #183, where the goal was to remove warnings from `-Wredundant-decls` flag. This is a nice goal, but redundant declarations aren't that harmful. `gcc` will issue an error if declared prototypes don't match, so these redundant function prototypes are *necessarily* the same declaration.

When removing the declaration from the header we would force every file that uses the `UNITY_OUTPUT_CHAR()` function to include a preprocessor guarded section. There are many existing users of Unity, and I didn't feel it was worth it to force code changes from this.

Examples:
***old way***
> #include "myOutput.h"
> #define UNITY_OUTPUT_CHAR myOutput
> #include "unity.h" // <- issues redundant declaration with default behavior

***now*** - the above example works without a warning if `redundant-decls` is requested, but...
> #include "unity.h"
> UNITY_OUTPUT_CHAR('0'); // <- issues an implicit declaration warning, which is worse
// ***unless*** - the user adds this preprocessor guarded declaration above
> #ifndef UNITY_DECL_ISNT_NEEDED // macro has a different name but same meaning
> int UNITY_OUTPUT_CHAR(int);
> #endif

***after this PR*** - back to old default behavior, with added config option to make the first example OK
> #include "myOutput.h"
> #define UNITY_OUTPUT_CHAR myOutput
> #include "unity.h" // <- redundant declaration can be suppressed if required by your code standard

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,6 @@ rvm:
   - "2.0.0"
 script:
   - cd test && rake ci
+  - make -s
+  - cd ../extras/fixture/test && rake ci
+  - make -s default noStdLibMalloc
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ script:
   - cd test && rake ci
   - make -s
   - cd ../extras/fixture/test && rake ci
-  - make -s default noStdLibMalloc
+  - make -s default noStdlibMalloc
--- a/src/unity.c
+++ b/src/unity.c
@@ -7,6 +7,9 @@
 #include "unity.h"
 #include <stddef.h>
 
+#ifdef UNITY_OMIT_OUTPUT_CHAR_HEADER_DECLARATION
+int UNITY_OUTPUT_CHAR(int); //If omitted from header, declare it here so it's ready for use
+#endif
 #define UNITY_FAIL_AND_BAIL   { Unity.CurrentTestFailed  = 1; longjmp(Unity.AbortFrame, 1); }
 #define UNITY_IGNORE_AND_BAIL { Unity.CurrentTestIgnored = 1; longjmp(Unity.AbortFrame, 1); }
 /// return prematurely if we are already in failure or ignore state
--- a/src/unity.c
+++ b/src/unity.c
@@ -7,11 +7,6 @@
 #include "unity.h"
 #include <stddef.h>
 
-#ifndef UNITY_OUTPUT_CHAR_USE_PUTC
-//If defined as something else, make sure we declare it here so it's ready for use
-extern int UNITY_OUTPUT_CHAR(int);
-#endif
-
 #define UNITY_FAIL_AND_BAIL   { Unity.CurrentTestFailed  = 1; longjmp(Unity.AbortFrame, 1); }
 #define UNITY_IGNORE_AND_BAIL { Unity.CurrentTestIgnored = 1; longjmp(Unity.AbortFrame, 1); }
 /// return prematurely if we are already in failure or ignore state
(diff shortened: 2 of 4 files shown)


xor-gate:
> This was changed in PR #183, where the goal was to remove warnings from -Wredundant-decls flag. This is a nice goal, but redundant declarations aren't that harmful. gcc will issue an error if declared prototypes don't match, so these redundant function prototypes are necessarily the same declaration.

What about compiling with `-Werror -Wredundant-decls` ? Redundant declaration ARE harmful because it only needs to be declared once (for reviewing). Even when ugly-auto-insert with the defines and preprocessor macros it should NOT produce redundant declarations.

rikvdh:
I agree with @xor-gate, having redundant declarations is generally wrong and useless. Within strict programming environments (Automotive, Aerospace) all odd programming quirks must be omitted. `-Werror -Wredundant-decls` is not uncommon in these businesses.

jsalling:
Here are my thoughts. I'm waiting for Mark to chime in on this one. @mvandervoord 

The main points here are:
* the Fixture build was broken by the previous change
* test code would be broken for existing users, too
* the change placed `redundant-decls` ahead of `implicit declaration` in order of importance

> What about compiling with -Werror -Wredundant-decls ?

I added an option for just this reason. Use: `UNITY_OMIT_OUTPUT_CHAR_HEADER_DECLARATION` Either at compile time, or `#define` it above `unity.h` to get rid of `redundant-decl` warnings. It is more customizable than a one-size-fits-all approach. Was this not clear?

> Redundant declaration ARE harmful

But not as harmful as implicit declarations; the warning is on by default. Code should avoid this first and foremost.
If you re-read the examples I provided, you can see I accounted for all of this and I tested it in these use cases.

> __rikvdh__ commented
> ...having redundant declarations is generally wrong and useless.

Generally useless, sure, but not at the expense of `implicit declaration` errors for existing users. Every user that uses `UNITY_OUTPUT_CHAR` will have this issue when overriding the macro name. *Useless*, but not wrong:
> MISRA C Rule 8.4: If functions are declared more than once their types shall be compatible.

Remember, folks use Unity in embedded systems and define output as `outbyte`, `uart_put`, `outChar`, etc. Unfortunately, the recent changes broke that behavior.

I get the desire to remove warnings. I have contributed a lot of code to Unity that does just that. I'd like to find something that works for all users of this project. I'm happy to add to this PR.

mvandervoord:
Sorry guys... been busy over here and am just catching up on this thread. This was a lot to soak in all at once. ;)

@rikvdh and @xor-gate have good points. I would definitely like to avoid this warning if possible. As @jsalling pointed out, though, the issue he is correcting is the more severe of the two. As an added bonus, he's fixing the situation where we are currently not actually running the fixture tests (yikes!). 

So I'm going to merge this pull request as the current best solution. If any of the three of you see a way to fix it so we don't have any of these issues, I'm sure a pull request which makes us all happy would be welcome!

Thank you all for chiming in here. It's been really nice having your help recently!

xor-gate:
Hi @jsalling, @mvandervoord thanks for catching this up. And for clearifying because it was not 100% clear for me at first sight what it would resolve. I hope I did not produce to much noise.

mvandervoord:
Not at all. It's great to have someone double-checking!

Issue #182, opened 2016-04-13
Redunant declarations compiler error for UnityPrintFail, UnityPrintOk and setUp


xor-gate:
Hi,

I'm using gcc arm embedded toolchain (2015q4, gcc 5.2) and have enabled `-Werror=redundant-decls` compiler flag. It seems the functions `UnityPrintFail`, `UnityPrintOk` and `setUp` have been defined to many times.
The definitions on line 70 and 71 and 1264 are not necessary and redundant.

Here is the full log:
```
Unity/src/unity.c:303:6: error: redundant redeclaration of 'UnityPrintFail' [-Werror=redundant-decls]
 void UnityPrintFail(void);
      ^
Unity/src/unity.c:70:6: note: previous declaration of 'UnityPrintFail' was here
 void UnityPrintFail(void);
      ^
Unity/src/unity.c:309:6: error: redundant redeclaration of 'UnityPrintOk' [-Werror=redundant-decls]
 void UnityPrintOk(void);
      ^
Unity/src/unity.c:71:6: note: previous declaration of 'UnityPrintOk' was here
 void UnityPrintOk(void);
      ^
Unity/src/unity.c:1264:10: error: redundant redeclaration of 'setUp' [-Werror=redundant-decls]
     void setUp(void);
          ^
In file included from Unity/src/unity.c:7:0:
Unity/src/unity.h:18:6: note: previous declaration of 'setUp' was here
 void setUp(void);
      ^
Unity/src/unity.c:1265:10: error: redundant redeclaration of 'tearDown' [-Werror=redundant-decls]
     void tearDown(void);
          ^
In file included from Unity/src/unity.c:7:0:
Unity/src/unity.h:19:6: note: previous declaration of 'tearDown' was here
 void tearDown(void);
```

It seems when I add this flag to the tests makefile it seems to fail also. It is a good idea to add this flag to the tests makefile when this is fixed. Using (debian 8 amd64):

```
cc --version
cc (Debian 4.9.2-10) 4.9.2
Copyright (C) 2014 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

```
jjacobs@dev03:~/repos/github/Unity/test$ make
cd build && \
[...]

PR #187, opened 2016-04-26
Followup of PR #185 without breaking the UNITY_OUTPUT_CHAR behaviour

xor-gate:
Fixture building was broken due to PR #182 and was not catched by travis. Now this is added and the output spy is declared extern for now because I have no clue how this is organized. This is probably a much better fix for now compared to PR #185

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,6 @@ rvm:
   - "2.0.0"
 script:
   - cd test && rake ci
+  - make -s
+  - cd ../extras/fixture/test && rake ci
+  - make -s default noStdlibMalloc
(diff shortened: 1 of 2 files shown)


jsalling:
I have to disagree that this is a better approach. Adding a specific declaration:
`int  UnityOutputCharSpy_OutputChar(int c)`  (due to an implicit declaration error during test) 
just to cover up this single use case is too narrow.
Remember, `UNITY_OUTPUT_CHAR` can be defined as anything.

I'm going to take this discussion back over to #185. I'd rather fix it up on the existing pull request.

xor-gate:
`UNITY_OUTPUT_CHAR` MAY NOT be anything because the implementer should match the strict prototype of `int putchar(int char)`. When the user/application overwrites UNITY_OUTPUT_CHAR the whole `unity.c` code where `UNITY_OUTPUT_CHAR` is used needs to know how the replaced macro looks like and needs to be extern inside `unity.c` because there is no way to signal the implementation of where the user/app defined function is declared. Actually the fixture building is broken and need to be fixed, not the other way around.

PR #192, opened 2016-05-05
100% code coverage for Unity project

jsalling:
This series of commits is mostly in service of getting code coverage for Unity. The new tests specifically cover lines that were untested before. Refactoring code simplifies things a little bit.

* Add test for failure count and UnityEnd return value
* Test number arrays comparison, with length equal to 0
* Remove weak linked setup/teardown from coverage build, unreachable & empty, no need to cover
* Simplify printing StrPointless into a single macro, remove repeated code
* Refactor UnityPrintNumber(), remove unreachable code. Remove what looks like an overflow check. The variable `divisor` cannot overflow.

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -491,6 +491,13 @@ void UnityAssertEqualNumber(const _U_SINT expected,
     }
 }
 
+#define UnityPrintPointlessAndBail()       \
+{                                          \
+    UnityTestResultsFailBegin(lineNumber); \
+    UnityPrint(UnityStrPointless);         \
+    UnityAddMsgIfSpecified(msg);           \
+    UNITY_FAIL_AND_BAIL; }
+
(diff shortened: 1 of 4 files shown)


mvandervoord:
REALLY nice work. Thanks so much for all the effort.

Before I merge it, I just had one really minor question. In c5238ad, what is the purpose of the test change on line 1373 of testunity.c? It seems like that was a good test before, and that it will (purposefully) fail for completely unrelated reasons now?

jsalling:
The main purpose there was to get coverage for `UnityPrintLen` printing escape characters. It was a bit of shorthand, I have to admit. The test right above it also tests `"bar\r"` against `"bar\n"` (fails the last character). This was the same test, in essence, but now it fails on character # 1. Hmm… I could have made the strings 5 characters long instead.

PR #193, opened 2016-05-11
Add template example unity_config.h file

wolf99:
Makes configuring unity quicker by having all configuration defines ready to uncomment.
Wasn't sure exactly which folder this should belong in, but if Unity is in use then ./src/ is already on the include path, if another folder would be better I can redo the PR.
Might go against Unity's minimalist footprint idea, if so please reject.


mvandervoord:
:) We've debated this very idea a number of times, actually. I like the way you pulled in the instructional portions with your file too.

Here's the reason we have so far resisted doing this: A lot of people are using Unity in their projects as a directory that they just pull in periodically to get the latest version. If those people are also using unity_config.h, they are left with either a merging headache or two instances of the file in their paths.

Maybe this means we include a file like this in an example directory instead? That way it would be accessible to everyone if needed but wouldn't be in the way for people who are already using a config header.

wolf99:
Good point re the include path @mvandervoord, I realise I would have the exact same issue!

Do you mean you'd like to have an example in the examples dir that *uses* this? Or just plonk it in the examples dir by itself?

mjago:
I like the helpful annotation provided too. Perhaps rake could supply an init command (similar to npm init (NodeJS package manager) that quickly walks you through the available options (complete with helpful messages) and builds a config file for you giving it a _specific name_ on completion. Perhaps the build system could attempt to pattern-match the generated config file and use it if it exists. This should be backward compatible and gitignore could ignore the pattern-match?

mvandervoord:
@wolf99 Either is good, if you feel strongly one way or the other. I was originally thinking on its own, but I see either is being fine.

@mjago This is something that would make sense to roll into Ceedling. It already has  a project init, which would be awesome to be a little more guided. For Unity itself, maybe a script in auto would make sense too.

wolf99:
Moved to examples by itself then by d7a371a9. WRT scripts and ruby, I'm afraid thats far from my wheelhouse at the moment. Maybe @mjago would be willing to add this?

mvandervoord:
It's really a much bigger thing than the scope of this PR, so we can always call that a new PR if we get to that point. :)

PR #199, opened 2016-06-21
Update unity_fixture_Test.c

bryongloden:
On line 530 deallocating a deallocated pointer: n1 is an error.

Found by https://github.com/danmar/cppcheck


bryongloden:
🎉

mvandervoord:
I believe your tool is wrong. If a realloc fails, it leaves the original pointer and memory intact. According to the standard: 

**7.20.3.4 The realloc function**

If ptr is a null pointer, the realloc function behaves like the malloc function for the specified size. Otherwise, if ptr does not match a pointer earlier returned by the calloc, malloc, or realloc function, or if the space has been deallocated by a call to the free or realloc function, the behavior is undefined. **If memory for the new object cannot be allocated, the old object is not deallocated and its value is unchanged.**

bryongloden:
@danmar 😁

PR #205, opened 2016-07-26
Update unity_fixture.c

bryongloden:
[[../Unity-master/extras/fixture/src/unity_fixture.c:210](https://github.com/ThrowTheSwitch/Unity/blob/master/extras/fixture/src/unity_fixture.c#L210)]: (error) Memory leak: guard

Found by [https://github.com/bryongloden/cppcheck](https://github.com/bryongloden/cppcheck)

### DIFF
--- a/extras/fixture/src/unity_fixture.c
+++ b/extras/fixture/src/unity_fixture.c
@@ -207,6 +207,7 @@ void* unity_malloc(size_t size)
     mem = (char*)&(guard[1]);
     memcpy(&mem[size], end, sizeof(end));
 
+    free(guard);
     return (void*)mem;
 }
 
--- a/extras/fixture/src/unity_fixture.c
+++ b/extras/fixture/src/unity_fixture.c
@@ -207,7 +207,9 @@ void* unity_malloc(size_t size)
     mem = (char*)&(guard[1]);
     memcpy(&mem[size], end, sizeof(end));
 
+#ifndef UNITY_EXCLUDE_STDLIB_MALLOC
     free(guard);
+#endif
     return (void*)mem;
 }
 
--- a/extras/fixture/src/unity_fixture.c
+++ b/extras/fixture/src/unity_fixture.c
@@ -207,7 +207,7 @@ void* unity_malloc(size_t size)
     mem = (char*)&(guard[1]);
     memcpy(&mem[size], end, sizeof(end));
 
-#ifndef UNITY_EXCLUDE_STDLIB_MALLOC
+#ifndef UNITY_FIXTURE_MALLOC_OVERRIDES_H_
     free(guard);
 #endif
     return (void*)mem;


mvandervoord:
This submission fails the automated self-tests. I suspect it's because it is only valid #ifndef UNITY_EXCLUDE_STDLIB_MALLOC?

bryongloden:
Nope! My tool is reporting the error when no tokens have been defined (but not when `UNITY_EXCLUDE_STDLIB_MALLOC`, is defined) i.e.

Checking [../Unity-master/extras/fixture/src/unity_fixture.c](https://github.com/ThrowTheSwitch/Unity/blob/master/extras/fixture/src/unity_fixture.c)...
[[../Unity-master/extras/fixture/src/unity_fixture.c:210](https://github.com/ThrowTheSwitch/Unity/blob/master/extras/fixture/src/unity_fixture.c#L210)]: (error) Memory leak: `guard`
Checking [../Unity-master/extras/fixture/src/unity_fixture.c](https://github.com/ThrowTheSwitch/Unity/blob/master/extras/fixture/src/unity_fixture.c): `UNITY_EXCLUDE_STDLIB_MALLOC`...
19/47 files checked 17% done

I suspect the error is related to [the command](https://travis-ci.org/ThrowTheSwitch/Unity/jobs/147550608#L693) `"make -s default noStdlibMalloc" exited with 2`. But because I'm unfamiliar with this command, do I close the PR?

bryongloden:
LOL. I believe you now! It IS related to UNITY_EXCLUDE_STDLIB_MALLOC. The answer was hidden inside the [Makefile](https://github.com/ThrowTheSwitch/Unity/blob/master/extras/fixture/test/Makefile#L36).

bryongloden:
Hi @mvandervoord ! I believe I edited [unity_fixture.c](https://github.com/ThrowTheSwitch/Unity/blob/master/extras/fixture/src/unity_fixture.c) correctly in https://github.com/ThrowTheSwitch/Unity/pull/205/commits/0737b4117fdc74a33a24080bd1f1a28069d4eb9b (by including `#ifndef` around my patch code). However, the new submission failed the automated self-tests for presumably the same reason. Any more hints before I close this and https://github.com/ThrowTheSwitch/Unity/pull/206?

mvandervoord:
It seems like it is disagreeing with your change altogether. The error you are getting at this point is "*** glibc detected *** build/framework_test.exe: double free or corruption (fasttop): 0x0000000001444010 ***"... I *think* this happens when release_memory is called, because you are now removing the guard object before then.

Why is it that you thought this was necessary? What situation are you covering for?

bryongloden:
Hi @mvandervoord. Thank you for looking over this. We've done studies you know; 60% of time [cppcheck](https://github.com/bryongloden/cppcheck) works all the [time](https://www.youtube.com/watch?v=pjvQFtlNQ-M) 👍.

Here we think it's necessary to free() memory because we believe its output. And [here](http://stackoverflow.com/a/4339459) [Steve Jessop](http://stackoverflow.com/users/13005/steve-jessop) enumerates the lists of situations we're 'covering' for (he even includes some situations we're not covering for 👍).

If you want/need a specific answer detailing the reasons our program detected an error in _your_ code, then I suspect it'll have to wait until after dinner (at the very least 👍).

Although, if you're confident in your implementation of memory management, then just close this and PR https://github.com/ThrowTheSwitch/Unity/pull/206 and chalk it all up as volunteerism 😁

mvandervoord:
I believe your tool has identified a problem as I look through this... I think your solution is wrong, though. I could be mistaken (this isn't actually my code... this file is an addon submitted by a third party. It's used by roughly a quarter of the Unity users, though, so I'd love to fix it up).

Anyway, your solution frees the memory before it's even called upon to be used. That's why you're running into an error about double-freeing memory.

In the most common case, I think the current system works. Memory is allocated in unity_malloc and it is freed again in unity_free. However, part of the point of this subsystem is to catch errors that people make, like not calling free when they should... and now that I am staring at it, I don't actually SEE a place where it checks to see if anyone ever came back and cleared that memory!

I think this is probably intended to happen somewhere in UnityConcludeFixtureTest?

Let me know what you think after dinner. ;)

bryongloden:
Commits https://github.com/ThrowTheSwitch/Unity/pull/205/commits/b1d8388ab85ac74f303b3384d25e805a56088a67 and https://github.com/ThrowTheSwitch/Unity/pull/205/commits/0737b4117fdc74a33a24080bd1f1a28069d4eb9b fixed with https://github.com/ThrowTheSwitch/Unity/pull/205/commits/4fd5f79f28562ac3e627bb5b4a590aaa32317902, thank you @mvandervoord

jsalling:
Oops. This tool is suggesting to free `guard`, but it should never be freed during `unity_malloc`.
The guard has the job of, well, _guarding_ the malloc'd memory passed to the caller. It's also how the implementation frees the total memory.

I'm going to revert this commit in another PR. The preprocessor conditional is never going to be true, so this is dead code, which is why it works.

Quote:
> However, part of the point of this subsystem is to catch errors that people make, like not calling free when they should... and now that I am staring at it, I don't actually SEE a place where it checks to see if anyone ever came back and cleared that memory!
I think this is probably intended to happen somewhere in UnityConcludeFixtureTest?

@mvandervoord This happens in the `UnityTestRunner`. Each test that uses this part of the fixture must `free` for each `malloc`, `calloc`, or `realloc`.

bryongloden:
Yes, unfortunately, sound and complete static analysis is shown to be undecidable. We're led to this false positive because we've adopted some unsound techniques in our tool.

PR #217, opened 2016-09-03
Use UNITY_FLOAT_INVALID_TRAIT instead of default.

2trill2spill:
Using default throws a warning when using `-Wall -Weverything` on clang. `UNITY_FLOAT_INVALID_TRAIT` was also the only enum value not explicitly handled in the switch statement, so handle it explicitly.

I noticed that the first commit failed on gcc because the switch does not have a `default:` case. So I added a `default:` and have it  `UNITY_FAIL_AND_BAIL;` to handle error cases. Do you think this is necessary? I usually use `default:` to handle values not in the enum in my code. But these commits fixes a clang warning for me, `-Wswitch-enum`. However it introduces a covered switch warning on clang, so these two commits probably are not worth it. I'm going to see if both compilers can be satisfied warning wise.

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -750,7 +750,7 @@ void UnityAssertFloatSpecial(const _UF actual,
                 is_trait = 1;
             break;
 
-        default:
+        case UNITY_FLOAT_INVALID_TRAIT:
             trait_index = 0;
             trait_names[0] = UnityStrInvalidFloatTrait;
             break;
--- a/src/unity.c
+++ b/src/unity.c
@@ -754,6 +754,12 @@ void UnityAssertFloatSpecial(const _UF actual,
             trait_index = 0;
             trait_names[0] = UnityStrInvalidFloatTrait;
             break;
+
+        default:
+           /* A value outside the enum was used for style so
+             just fail because something has gone wrong. */
+            UNITY_FAIL_AND_BAIL;
+
     }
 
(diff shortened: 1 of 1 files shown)


2trill2spill:
It looks like the way you had it was probably best. It seems the options are...

1) Leave it the way it is.

2) Remove the `default:` case and add a case for `UNITY_FLOAT_INVALID_TRAIT`, but you would have to remove `-Wswitch-default` from your test runner build so that does not fail.

3) Use these two commits which explicitly handle `UNITY_FLOAT_INVALID_TRAIT` and change the `default:` case to using `UNITY_FAIL_AND_BAIL;` so it bails if a value not in the enum is passed to the switch statement. These commit also get rid of a clang warning but also adds a clang warning so it's a wash.

Considering all of those options It's probably best to just leave it the way it is, but let me know what you think. On a slightly related note it would be nice to add clang to your travis build. I will probably add a pull request to add osx to your travis config so you can test with clang as well.

jsalling:
Thanks for looking into this. You can read more history in #177. 
It is impossible to get rid of all warnings with `-Weverything`, so we agreed on a style choice.

I like your idea of expanding the Travis build. A Pull Request for that would be great.

2trill2spill:
Yea I think your right about just leaving it as is. I will open a pull request for the travis build when It's done.

PR #221, opened 2016-09-21
Test generator: scrub strings after comments

### LINKED ISSUE #220: Test runner generator ignores code across multiple lines between quotation marks
If your test file contains a line with an odd number of quotation marks and another line somewhere after it with at least one quotation mark, the [test runner generator](https://github.com/ThrowTheSwitch/Unity/blob/f276c79/auto/generate_test_runner.rb#L98) will ignore everything between the last quotation mark on the first line and the first quotation mark on the second line.

Example test file excerpt:
```c
void test_foo(void)
{
    unsigned char buf[] = {
        0x01, 0x02, 0x03, 0x22, 0x05, 0x06, 0x07, 0x08, /* ...".... */
    };
}

void test_bar(void)
{
    unsigned char buf[] = {
        0x01, 0x02, 0x03, 0x22, 0x05, 0x06, 0x07, 0x08, /* ...".... */
    };
} 
```

Because of this bug, the function `test_bar` will be ignored.

This bug was caused by commit bcf651532.

robobrobro:
This fixes #220.

Removing strings from test files is still dangerous, but much
less likely to cause problems after this change to do the
removal after removing comments.

The bug could still manifest if a test file contains defines two
macros, one that contains a single quotation mark and then another
defined somewhere after it that contains a single quotation mark.
Everything in between the aforementioned quotation marks would
still be ignored after this commit, but that is an unlikely
scenario.

### DIFF
--- a/auto/generate_test_runner.rb
+++ b/auto/generate_test_runner.rb
@@ -95,9 +95,9 @@ class UnityTestRunnerGenerator
     tests_and_line_numbers = []
 
     source_scrubbed = source.clone
-    source_scrubbed = source_scrubbed.gsub(/"[^"]*"/, '')      # remove things in strings
     source_scrubbed = source_scrubbed.gsub(/\/\/.*$/, '')      # remove line comments
     source_scrubbed = source_scrubbed.gsub(/\/\*.*?\*\//m, '') # remove block comments
+    source_scrubbed = source_scrubbed.gsub(/"[^"]*"/, '')      # remove things in strings
     lines = source_scrubbed.split(/(^\s*\#.*$)                 # Treat preprocessor directives as a logical line
                               | (;|\{|\}) /x)                  # Match ;, {, and } as end of lines
 
--- a/auto/generate_test_runner.rb
+++ b/auto/generate_test_runner.rb
@@ -95,9 +95,9 @@ class UnityTestRunnerGenerator
     tests_and_line_numbers = []
 
     source_scrubbed = source.clone
+    source_scrubbed = source_scrubbed.gsub(/"[^"\n]*"/, '')      # remove things in strings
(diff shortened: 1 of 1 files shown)


hbowden:
It looks like this commit breaks some tests on Linux and potentially on Mac OSX as well. Check out the travis-ci build output to see what tests are broken: https://travis-ci.org/ThrowTheSwitch/Unity/jobs/161711534

robobrobro:
I've realized that, while this fixes #220, it also brings to light the fact that commit bcf6515 causes other problems. As evidenced by the failed CI build, changing the order of string scrubbing in `find_tests` allows another bug in that function to manifest, which is that blocks that resemble multi-line C comments (`/* ... */`) are ruthlessly ignored by the test generator. See [test/testdata/testRunnerGenerator.c](https://github.com/ThrowTheSwitch/Unity/blob/693169e/test/testdata/testRunnerGenerator.c#L115) for an example of a string that contains the beginning of a C block comment `/*` and a later strsing that contains the end of a C block comment `*/`, which cause the `find_tests` to ignore the test in between, `test_NotDisappearJustBecaseTheTestBeforeAndAfterHaveCrazyStrings`.

The code in `find_tests()` is very error-prone.

robobrobro:
@hbowden What do you think about changing the regular expression for strings to `/"[^"\n]*'/` instead of moving the string scrubbing after the comment scrubbing?

robobrobro:
@hbowden Well, the GCC tests are working, but the OSX tests won't start, but I'm assuming those would be fine as well.

hbowden:
Travis CI for OSX has been terrible  the last few weeks, I'm going to open an issue with travis ci and complain about it :) . But I tested your changes locally on OSX and it passes all the tests.

robobrobro:
@hbowden Great. Thanks! I really appreciate your responsiveness :).

PR #231, opened 2016-11-24
Unity prints float (and double) by default

jsalling:
Unity Prints Floating Point By Default
When a floating point test assert fails the output now prints a number without enabling printing support.
* Breaks the dependency on `snprintf()`
* Uses a custom algorithm with a fixed format, similar to previous format (to six decimal places)
* Large numbers are printed in exponential form with 9 digits of precision: 1.8e+nn
* Examples: 12.654321, 1.25e+10, 1.0, -4100200300.0, NaN, -Inf
* Trims extra trailing 0's for human readable format, 2.0 vs. 2.000000 before
* Special indication for small numbers, 0.000000… vs. 0.0 before when < 5e-7 but not 0
* Twice as fast as `snprintf` in benchmarks, not that it matters for the occasional failure
* Numerical analysis shows correctness for all floats, unique values 16.0 and up
* A very small percent of doubles may round wrong in the last digit due to double precision arithmetic
* Rounds numbers to nearest value, and ties round to even (as is true for most floating point)
* Option to round ties away from 0, `UNITY_ROUND_TIES_AWAY_FROM_ZERO`
* Remove `UNITY_FLOAT_VERBOSE`, add option `UNITY_EXCLUDE_FLOAT_PRINT`
* Tests. Lots of corner cases and an optional test set running through all `float` values

### DIFF
--- a/auto/generate_module.rb
+++ b/auto/generate_module.rb
@@ -10,6 +10,7 @@
 
 require 'rubygems'
 require 'fileutils'
+require 'pathname'
 
 #TEMPLATE_TST
 TEMPLATE_TST ||= %q[#include "unity.h"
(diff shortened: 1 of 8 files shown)


mvandervoord:
This is awesome. 

Unfortunately we have stepped on eachother's toes. I just did a renaming of the internal variables. Things like this:

_UF -> UNITY_FLOAT
_UD -> UNITY_DOUBLE
_UU -> UNITY_UINT
_UU32 -> UNITY_UINT32
_US16 -> UNITY_INT16

I suspect it's mostly the first two that affect your latest changes. I can probably tackle this next week if you need me to. Otherwise, if you have time to do the cleanup earlier, that would be awesome.

jsalling:
Thanks for the kind words. I'm really glad you like it and think it's awesome.

I'll have to take you up on your offer to tackle this merge. I won't be able to do it.
I see there are merge conflicts now, but it doesn't show what they are. I think you're right about type name changes being the cause here. During the merge, the name conflicts can be sorted out. The types inside any new code will have to be renamed for the build to succeed, too.

Thanks again

jsalling:
@mvandervoord 
Pinging you on this PR. All conflicts fixed. Have you had a chance to review it?

mvandervoord:
Thanks so much for sticking with that one! I know it wasn't easy! Also, thanks for the ping!

PR #233, opened 2016-12-02
Module generator finishes for partially existing files

### LINKED ISSUE #219: Module Generator should still finish for partially existing files
For example, if I have meh.c and meh.h in my project, and I call the script to generator a module for meh, it should still add the Testmeh.h.

mchernosky:
This resolves #219. When generating a new module, if all the files to generate already exist then it fails as before. If some of the files already exist, then the files that need to be created are created. Any
existing files are not changed.

Also added a bunch of tests for this feature via rspec. Run them from the test folder with `rake spec`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build/
+test/sandbox
 .DS_Store
 examples/example_1/test1.out
 examples/example_1/test2.out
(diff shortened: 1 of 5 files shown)


mchernosky:
Er.. looks like the build broke because my tests need rspec.

mchernosky:
Updated the Travis configuration to install rspec, so the tests pass now.

PR #238, opened 2016-12-27
Refactor UnityAssertEqualIntArray

jsalling:
This refactor removes repeated code and makes the compiled size of `UnityAssertEqualIntArray` about 30% smaller. Essentially no functional changes. The cast-to-int to do pointer math is removed, as is the type UNITY_PTR which served in this role. All pointer math is done with pointer types now.


jsalling:
@mvandervoord 
Pinging you on PR 238

mvandervoord:
overall I like this cleanup. It's closer to what we had originally, but fixes many of the issues we had with that version.

One concern I see looking through it: You do numerous operations between a pointer and an integer without any casts. First, there is no guarantee that an int is large enough to hold a pointer. On many platforms it isn't. Second, some compilers may give you warnings or errors for this. Like this line:

expected = length + (const char*)expected;

expected is a void* pointer. length is an unsigned signed integer. Ideally they will always work together... but this function got as complicated as it is because that wasn't always true.

I'm tempted to merge and see who complains... but I have been hoping to look through it one more time for any glaring issues.

jsalling:
I don't think those issues are still present. Let me explain.
The variable `expected` is a `void*`, which indeed cannot do pointer arithmetic. The cast to `(const char*)` type-puns `expected` into a pointer with a fully-known type, `char`. Pointer arithmetic is allowed on a `char*` without a warning on any compiler that I know. Additionally, even though a pointer and `int` may not be the same size, arithmetic should be allowed between them with widening promotion, if required.
In C, you can spell pointer arithmetic with address-of and array operators, but I feel the syntax ends up being tricky to read in this case.
```C
// I prefer
expected = length + (const char*)expected;
// over
expected = &((const char*)expected)[length];
```
Let me know if I missed anything. What do you think?

Issue #246, opened 2016-12-31
Integer masking causes poor failure message


jsalling:
With 64-bit support on, example prints:
`FAIL: Expected 1 Was 1`

```C
#include "unity.h"
static void testOne(void)
{
    TEST_ASSERT_EQUAL_UINT(1, 4294967297);
}
int main(void)
{
    UNITY_BEGIN();
    RUN_TEST(testOne);
    return UNITY_END();
}
```

Clearly this is a problem. Masking before printing is the cause. The masking was mentioned recently in issue #235.

I propose eliminating the masking of UINT values before printing, in `UnityPrintNumberByStyle`. The failure message would then make sense.

PR #249, opened 2017-01-17
No print masking of integers, bounds on hex printing

### LINKED ISSUE #235: Potential buffer overrun in unity.c line 160
Line 160:

`UnityPrintNumberUnsigned(  (_U_UINT)number  &  UnitySizeMask[((_U_UINT)style & (_U_UINT)0x0F) - 1]  );`

If you assume that someone could pass in a style not based on the definitions in unity_internals.h then `(style & 0x0F) - 1` could result in an index of 14 into `UnitySizeMask[]` which only has 7 elements (or 8 with 64 bit support).

jsalling:
Closes #235 and #246
Masking of numbers could be useful, but currently causes bad failure messages as in #246.
1) There is asymmetry in the masks. The print function only masks unsigned values, but fails to mask signed values.
2) All the required masking is done by casts in the calling macro. This bitwise masking may be vestigial.
3) Upper bounds prevent printing more hex characters than the number has. Only useful if a bad length was passed into a print function.

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -190,7 +190,8 @@ void UnityPrintNumberHex(const UNITY_UINT number, const char nibbles_to_print)
 {
     int nibble;
     char nibbles = nibbles_to_print;
-    if ((unsigned)nibbles > 2 * sizeof number) nibbles = 2 * sizeof number;
+    if ((unsigned)nibbles > (2 * sizeof(number)))
+        nibbles = 2 * sizeof(number);
 
     while (nibbles > 0)
     {
@@ -198,11 +199,11 @@ void UnityPrintNumberHex(const UNITY_UINT number, const char nibbles_to_print)
         nibble = (number >> (nibbles * 4)) & 0x0F;
         if (nibble <= 9)
         {
-            UNITY_OUTPUT_CHAR('0' + nibble);
+            UNITY_OUTPUT_CHAR((char)('0' + nibble));
         }
         else
         {
-            UNITY_OUTPUT_CHAR('A' - 10 + nibble);
+            UNITY_OUTPUT_CHAR((char)('A' - 10 + nibble));
         }
     }
 }
--- a/src/unity.c
+++ b/src/unity.c
@@ -52,21 +52,6 @@ static const char UnityStrResultsIgnored[]         = " Ignored ";
 static const char UnityStrDetail1Name[]            = UNITY_DETAIL1_NAME " ";
 static const char UnityStrDetail2Name[]            = " " UNITY_DETAIL2_NAME " ";
 
-/* compiler-generic print formatting masks */
-static const UNITY_UINT UnitySizeMask[] =
-{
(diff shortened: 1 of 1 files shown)


src/unity.c
```diff
@@ -203,19 +188,21 @@ void UnityPrintNumberUnsigned(const UNITY_UINT number)
...
+        nibble = (number >> (nibbles * 4)) & 0x0F;
         if (nibble <= 9)
         {
-            UNITY_OUTPUT_CHAR((char)('0' + nibble));
+            UNITY_OUTPUT_CHAR('0' + nibble);
         }
         else
         {
-            UNITY_OUTPUT_CHAR((char)('A' - 10 + nibble));
+            UNITY_OUTPUT_CHAR('A' - 10 + nibble);
```
mvandervoord: The cast here is important to avoid the "possible loss of data in implicit cast" warning

src/unity.c
```diff
@@ -203,19 +188,21 @@ void UnityPrintNumberUnsigned(const UNITY_UINT number)
 /*-----------------------------------------------*/
 void UnityPrintNumberHex(const UNITY_UINT number, const char nibbles_to_print)
 {
-    UNITY_UINT nibble;
+    int nibble;
     char nibbles = nibbles_to_print;
+    if ((unsigned)nibbles > 2 * sizeof number) nibbles = 2 * sizeof number;
```
mvandervoord: sizeof(number) is standard. This notation is not.

mvandervoord: plus, we try to go overboard with parenthesis on this project:

if ((unsigned) nibbles > (2 * sizeof(number)) )
    nibbles = 2 * sizeof(number);

src/unity.c
```diff
@@ -203,19 +188,21 @@ void UnityPrintNumberUnsigned(const UNITY_UINT number)
 /*-----------------------------------------------*/
 void UnityPrintNumberHex(const UNITY_UINT number, const char nibbles_to_print)
 {
-    UNITY_UINT nibble;
+    int nibble;
```
mvandervoord: why not UNITY_INT?

jsalling: Just doing the simplest thing... 
I've learned `UNITY_INT` is always 32-bits or 64-bits, and `int` matches the register size of the target platform. The value assigned is bounded in the range {0,15}, so any type could work. The sum below, `'0' + nibble`, becomes an int, too.

src/unity.c
jsalling: I'll follow your lead on this - happy to change it.

src/unity.c
jsalling: ...changing it back.

What I was thinking:
Since the prototype is:
```C
void UNITY_OUTPUT_CHAR(int);
```
I guessed that the narrowing cast wasn't required. I do test these things with several compilers, but it's easy to miss a warning. Thanks for catching it.

Issue #181, opened 2016-04-11
TEST_PASS() doesn't play well with UNITY_EXCLUDE_SETJMP_H=1


l3nticular:
Commit 433ee635755457f0cc428826b3967034095224cd introduced this macro:
#define TEST_PASS()                                                                                longjmp(Unity.AbortFrame, 1)

and this function:

void testPassShouldEndImmediatelyWithPass(void)
{
    TEST_PASS();
    TEST_FAIL_MESSAGE("We should have passed already and finished this test");
}

However, on a platform without setjmp/longjmp, this can't work properly since there are no ifdefs around either of these.  On a platform with out this, it should probably alias to TEST_IGNORE_MESSAGE(,"Not supported") or similar.

l3nticular:
Ignore me....

PR #250, opened 2017-01-17
Optional UNITY_EXCLUDE_SETJMP_H, different control flow

jsalling:
This option requires replacing all `longjmp` calls with `return`. Unity's control flow is different if a test fails **and** there is more code after the failing line. Instead of exiting immediately to `tearDown()`, the rest of the test executes first.
Since this is targeted for very constrained systems, the compromise seems realistic. It keeps this feature very simple, too. One can write tests to avoid side-effects, as appropriate.
Passing tests behave exactly the same.
A feature like this was brought up in the closed issue #181.

* Condense all `longjmp` calls into TEST_ABORT()
The `setjmp` calls are all in TEST_PROTECT() already
* Rename Array Check helper, always `return`, never `longjmp`
Move longjump inside caller, to functions returning `void`
This single function needed to change to allow optional `setjmp.h`
* Cleanup Assert Array idiom to be the same everywhere
* Option to exclude setjump/longjmp, `setjmp.h`
Using this option changes the control flow of Unity, but is useful on
constrained embedded systems. You can't fully simulate the power of
`longjmp` with just `return`, but Unity still works well, all tests pass.

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -480,32 +480,32 @@ static void UnityPrintExpectedAndActualStringsLen(const char* expected, const ch
  * Assertion & Control Helpers
  *-----------------------------------------------*/
 
-static int UnityCheckArraysForNull(UNITY_INTERNAL_PTR expected, UNITY_INTERNAL_PTR actual, const UNITY_LINE_TYPE lineNumber, const char* msg)
+static int UnityIsOneArrayNull(UNITY_INTERNAL_PTR expected,
+                               UNITY_INTERNAL_PTR actual,
+                               const UNITY_LINE_TYPE lineNumber,
+                               const char* msg)
 {
-    /* return true if they are both NULL */
-    if ((expected == NULL) && (actual == NULL))
-        return 1;
+    if (expected == actual) return 0; /* Both are NULL or same pointer */
 
-    /* throw error if just expected is NULL */
+    /* print and return true if just expected is NULL */
     if (expected == NULL)
     {
         UnityTestResultsFailBegin(lineNumber);
         UnityPrint(UnityStrNullPointerForExpected);
         UnityAddMsgIfSpecified(msg);
-        UNITY_FAIL_AND_BAIL;
+        return 1;
     }
 
-    /* throw error if just actual is NULL */
+    /* print and return true if just actual is NULL */
     if (actual == NULL)
     {
         UnityTestResultsFailBegin(lineNumber);
         UnityPrint(UnityStrNullPointerForActual);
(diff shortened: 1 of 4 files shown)


src/unity.c
```diff
@@ -573,10 +573,7 @@ void UnityAssertEqualIntArray(UNITY_INTERNAL_PTR expected,
 
     RETURN_IF_FAIL_OR_IGNORE;
 
-    if (num_elements == 0)
```
mvandervoord: Our "coding standard" (in quotes, because clearly we don't do a good job of telling people what it is) is to always have braces and to always have the THEN portion of an if statement on a newline.

src/unity.c
```diff
@@ -676,33 +673,28 @@ void UnityAssertEqualFloatArray(UNITY_PTR_ATTRIBUTE const UNITY_FLOAT* expected,
                                 const UNITY_LINE_TYPE lineNumber)
 {
     UNITY_UINT32 elements = num_elements;
-    UNITY_PTR_ATTRIBUTE const UNITY_FLOAT* ptr_expected = expected;
-    UNITY_PTR_ATTRIBUTE const UNITY_FLOAT* ptr_actual = actual;
```
mvandervoord: This additional assignment is here to make the PIC compilers happy. They require a particular attribute for some of the compilers to be assigned to the pointers, letting the compiler no they are a non-standard size.

src/unity.c
```diff
@@ -13,8 +13,8 @@ void UNITY_OUTPUT_CHAR(int);
 #endif
 
 /* Helpful macros for us to use here in Assert functions */
-#define UNITY_FAIL_AND_BAIL   { Unity.CurrentTestFailed  = 1; longjmp(Unity.AbortFrame, 1); }
-#define UNITY_IGNORE_AND_BAIL { Unity.CurrentTestIgnored = 1; longjmp(Unity.AbortFrame, 1); }
+#define UNITY_FAIL_AND_BAIL   { Unity.CurrentTestFailed  = 1; TEST_ABORT(); }
```
mvandervoord: I like this. it's much cleaner

src/unity.c
jsalling: I'll follow your lead on the style.
I think I can just revert the Cleanup commit.

src/unity.c
jsalling: Thanks

src/unity.c
jsalling: Good to know. Reverting...

mvandervoord:
Thanks! You rock!

PR #252, opened 2017-01-19
Compile examples during continuous integration

### LINKED ISSUE #216: Examples should compile during TravisCI build
Unity comes with 3 examples. Compiling them during continuous integration will help keep them up to date.

jsalling:
Closes #216
* Get example 1 to compile easily on systems without ruby
* Cleanup example Makefiles, update test runners
* Compile examples in continuous integration, update example_3 readme
Altered the rake build so that it passes. The example_3 shows off
some failing tests, so the rake build has to ignore those.
Update .gitignore file with executables from examples


jsalling:
Wow, apparently the Travis-CI Linux had issues with the 32-bit build and failed on the first try.
I patched it: adding an install of `gcc-multilib` which gives the compiler what it wants for 32-bit builds on Linux.
Clang on OSX was smooth sailing.

examples/example_2/makefile
```diff
@@ -5,31 +5,31 @@
...
+C_COMPILER=clang
+endif
+
+UNITY_ROOT=../..
 
-CFLAGS = -std=c99
+CFLAGS=-std=c99
 CFLAGS += -Wall
 CFLAGS += -Wextra
-CFLAGS += -Werror
```
jsalling: The examples compile without any warnings on most versions of gcc, nor did they throw warnings before. I also removed a flag that was for C++, and checked the two examples have the same warning flags.

I removed `-Werror` specifically due to how warnings change in compilers over time. I recently found out that gcc 6.1 can issue a warning for inconsistent & misleading indentation. It's even on by default. This applies here, since the example uses misleading indentation in the code. (An example of unit testing catching mistakes)
If `-Werror` is enabled, up-to-date users will download the example and find it doesn't build at all.
Philosophically, this compromise felt better for users, to me at least.
I wasn't ready to change up the whole example. Thoughts?

Also, this branch is open to edits from maintainers. You can commit to it and make changes, if you'd like.

jsalling:
I had an idea. How about build continuous integration with `-Werror` but leave the default target as-is?
This catches more issues in CI and leaves new users with an example build less likely to fail.
`make ci`
I pushed a new commit.

Issue #256, opened 2017-01-24
Mismatched declaration and definition for UnityPrintFloat causes compiler warnings


dmcminniwc:
In unity.c UnityPrintFloat is defined as  `void UnityPrintFloat(UNITY_DOUBLE number)`  but it is declared in unity_internals.h as `void UnityPrintFloat(const UNITY_DOUBLE number)`. The parameter is marked as `const` in the header and not in the source which results in a compile time warning from MSVC at least. Please use const in the source as well as the header.

dmcminniwc:
Leave a comment

PR #257, opened 2017-01-24
Align definition and declaration of UnityPrintFloat

Letme:
As reported in #256 there is a mismatch between declaration and definition of UnitPrintFloat. Removing const from the parameter, so that definition and declaration match.


Letme:
Then I rather remove the const in that case than waste a variable. There is only one location where we write to number :disappointed:

mvandervoord:
I'm not sure I agree. By setting it to const, we allow people to test variables that are floats and const floats. I think this is worth burning a couple bytes of stack space.

mvandervoord:
Awesome. :)

PR #264, opened 2017-02-23
Added "less / greater than" comparison tests

Dommar92:
New Asserts:
TEST_ASSERT_GREATER_THAN_INT(shouldBeGreater, actual)
Success, if shouldBeGreater > actual, fails otherwise.

TEST_ASSERT_LESS_THAN_INT(shouldBeSmaller, actual)
Success, if shouldBeSmaller < actual, fails otherwise.

Needed this in a project, so I added it. May be useful to others.

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -7,446 +7,418 @@
 #include "unity.h"
 #include <stddef.h>
 
-/* If omitted from header, declare overrideable prototypes here so they're ready for use */
+/* If omitted from header, declare overrideable prototypes here so they're ready
+ * for use */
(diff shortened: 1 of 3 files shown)


art-of-dom:
While I like this change a lot, can you revert all the spacing/style changes in unity.c so it doesn't change as much. This will cause conflicts with forked and branched changes trying to merge back in as your changes are so big.

Dommar92:
Yes, you are right. Sorry for that! Will do it tomorrow!

mvandervoord:
If anyone would like to clean up this PR, I'd still be interested in merging it.

mvandervoord:
Merged the cleaner #285. Thanks for the work, everyone!

PR #272, opened 2017-03-18
Fix -Wconversion with gcc-4.3

FSMaxB:


src/unity.c
```diff
@@ -898,16 +898,16 @@ void UnityAssertNumbersWithin(const UNITY_UINT delta,
     if ((style & UNITY_DISPLAY_RANGE_INT) == UNITY_DISPLAY_RANGE_INT)
     {
         if (actual > expected)
-            Unity.CurrentTestFailed = ((UNITY_UINT)(actual - expected) > delta);
+          Unity.CurrentTestFailed = (UNITY_UINT)((UNITY_UINT)(actual - expected) > delta);
```
FSMaxB: Dammit, looks like I broke the indentation of this line.

mvandervoord: No problem. I'm fixing all sorts of stuff today. I can clean that up as I go.

Issue #273, opened 2017-03-20
Convert documentation to a structured text format instead of binary


wolf99:
Comment https://github.com/ThrowTheSwitch/Unity/issues/194#issuecomment-287890661 mentions the possibility of converting documentation to some structured text format instead of the current binary format (PDF).

I would be willing to put in leg work on this but wanted to discuss the most suitable format as I am mainly only experienced with markdown of those mentioned in the linked issue (markdown, restructured text, aciidoc, etc). I currently don't see any benefit to one of those over the others but I may well be wrong!

One consideration may be the ease with which PDF's could subsequently be programmatically generated from the resulting documents, if such a thing was required (but the merits of such a requirement could also be discussed).

wolf99:
Another consideration might be if some of the documentation might be able to be automatically from code comments (i.e. the work to then be done her would be to break down the current documentation fitting this, into code comments).

mvandervoord:
Do you know of any projects that have done this well? I have been disappointed by most of the automatically generated docs for the projects I've seen.

bjones1:
To advertise my own (free - GPL) tools a bit, [CodeChat](http://pythonhosted.org/CodeChat/) transforms reStructuredText comment in source code to web pages using Sphinx, which produces beautiful documentation directly from the source code -- [here's](http://pythonhosted.org/CodeChat/CodeChat/CodeToRest.py.html) an example. It supports C/C++, Ruby, and many more languages.

wolf99:
After a quick whiz around Google, on the point of generation from inline documentation in source code I can see two other options that jump out:

- Sphinx, uses restructured text, documents C and Ruby via 'Domains'  (I have zero experience with Sphinx or rst)
- Doxygen, uses markdown, doesn't document Ruby (But I have some little experience with this)

It is true however (at least in my experience of Doxygen and my quick perusal of doc gen websites this morning), that generating beautiful PDFs from takes, at the very least, considerable configuration twiddling. HTML and textual output is quite a bit easier though.

Letme:
I can add to this discussion. I have been using sphinx for quite some time now to convert reStructuredText to html document, but it also supports .pdf. What is amazing about Sphinx is the amount of plugins -> from plantuml to various plugins for whatever you might imagine. Source is similar to markdown and nice to diff in merge/pull requests. There are also some tools to convert doc to rst which work fine without any pictures or graphs (which I doubt I seen in current docs).

[...]

PR #276, opened 2017-03-29
Convert PDF docs to markdown

wolf99:
This is in aid of #273 as [requested](https://github.com/ThrowTheSwitch/Unity/issues/273#issuecomment-289619879)

Note that the conversion is direct as I could make it, thus errors in the source PDFs, such as broken or incorrectly formatted links, remain in the markdown version (mainly because I wasn't sure what to replace them with, so maybe they should constitute a new issue).


wolf99:
Not sure why the deletion of this particular PDF is causing a conflict? All PDFs that were deleted were deleted together and committed in a single commit. and my branch is currently from master where master is at its most recent.

mvandervoord:
I made a change to the AssertionsReference to add documentation for the new array alternatives TEST_ASSERT_EACH_EQUAL_INT and friends. It appears from the document that you generated, that your source was from before these changes were inserted. I suspect this is why there is a conflict. 

We can go with your version and then I'll just re-add those details. :)

wolf99:
Almost forgot to note that the markdown uses _Github-flavoured_ markdown rather than regular markdown. In a lot of cases this will not matter but some strict renderers may barf some things in it, such as the code blocks that use the three-back-tick fences, for example.

wolf99:
Ah, I see I pulled and rebased on _my_ remote, rather than the ThrowTheSwitch remote, D'oh! Ill rebase properly now

wolf99:
the branch is now rebased correctly and I have added the EACH_EQUAL changes 😃

PR #278, opened 2017-04-10
Naive implementation of printing execution time for each test

shreyasbharath:


shreyasbharath:
@mvandervoord Just an early look, feel free to comment.

extras/fixture/src/unity_fixture.c
```diff
@@ -7,7 +7,9 @@
 
 #include "unity_fixture.h"
 #include "unity_internals.h"
+#include <stdio.h>
```
shreyasbharath: I suppose we can't #include this directly?

extras/fixture/src/unity_fixture.c
```diff
@@ -7,7 +7,9 @@
...
 #include <string.h>
+#include <time.h>
```
shreyasbharath: Same with this one?

extras/fixture/src/unity_fixture.c
```diff
@@ -93,6 +95,8 @@ void UnityTestRunner(unityfunction* setup,
         UnityMalloc_StartTest();
         UnityPointer_Init();
 
+        Unity.CurrentTestStartTime = clock();
```
shreyasbharath: Not sure if this is the best way to capture the start time, but seemed sensible.

src/unity_internals.h
```diff
@@ -387,6 +387,7 @@ struct UNITY_STORAGE_T
     UNITY_COUNTER_TYPE TestIgnores;
     UNITY_COUNTER_TYPE CurrentTestFailed;
     UNITY_COUNTER_TYPE CurrentTestIgnored;
+    UNITY_COUNTER_TYPE CurrentTestStartTime;
```
shreyasbharath: Probably want to wrap this with a #ifdef?

extras/fixture/src/unity_fixture.c
```diff
@@ -417,7 +421,11 @@ void UnityConcludeFixtureTest(void)
     {
         if (UnityFixture.Verbose)
         {
-            UnityPrint(" PASS");
+            clock_t end = clock();
```
shreyasbharath: Instead of using clock() we would use a #define'd macro I guess?

shreyasbharath:
Here's example output from 'rake all'

execution_time_unity

extras/fixture/src/unity_fixture.c
shreyasbharath: Yup I realised I can use UnityPrintNumberUnsigned, so I can now remove this #include.

extras/fixture/src/unity_fixture.c
shreyasbharath: Yup makes sense, will change.

extras/fixture/src/unity_fixture.c
```diff
@@ -417,7 +421,11 @@ void UnityConcludeFixtureTest(void)
...
+            uint32_t timeTakenMs = ((end - Unity.CurrentTestStartTime) / (double)CLOCKS_PER_SEC * 1000.0f);
+            char msg[64];
+            snprintf(msg, sizeof(msg), " PASS [%u ms]", timeTakenMs);
```
shreyasbharath: Done.

shreyasbharath:
@mvandervoord Unit tests? Where should I add them?

extras/fixture/src/unity_fixture.c
```diff
@@ -93,6 +93,8 @@ void UnityTestRunner(unityfunction* setup,
         UnityMalloc_StartTest();
         UnityPointer_Init();
 
+        Unity.CurrentTestStartTime = CLOCK_MS();
```
shreyasbharath: Do we want to wrap this and the block below with the #ifdef too? Makes sense.

src/unity_internals.h
```diff
@@ -32,6 +32,10 @@
 #include <limits.h>
 #endif
 
+#ifndef UNITY_EXCLUDE_TIME_H
```
shreyasbharath: Open to a better name...

shreyasbharath:
@mvandervoord Any progress on this?

mvandervoord:
I like the overall idea. There are many embedded developers who aren't going to be able to use this feature. So, I think it should be only enabled with something like UNITY_INCLUDE_EXEC_TIME. 

I would push all the tracking and printing into a set of macros like UNITY_EXEC_TIME_START(), UNITY_EXEC_TIME_STOP() and UNITY_PRINT_EXEC_TIME(). Those can just be called directly in the Unity code. They'll equate to nothing most of the time, but when UNITY_INCLUDE_EXEC_TIME is included, they'll be real code that does something. 

I'd still keep your CLOCK_MS macro so the user can configure how time is determined. I would rename it to UNITY_CLOCK_MS for easy sandboxing. It can have a nice default value when UNITY_INCLUDE_EXEC_TIME is enabled so on systems support time.h it can just be used.

Sorry to dump all these changes, but I think we'll be much happier with the implementation if we keep these things optional. Believe me, I've been down this road on a number of other features already. ;)

Thanks for the help!!!

shreyasbharath:
@mvandervoord I have implemented the changes you suggested. Let me know if you'd like me to tidy it up even further.

Also, do you have any suggestions on how to unit test this?

shreyasbharath:
@mvandervoord Can you take a look and let me know what needs to be done to merge this?

mvandervoord:
You can see what is failing on Travis CI [here](https://travis-ci.org/ThrowTheSwitch/Unity/jobs/273741897). It appears that the problems are all related to new functions not having prototypes?

bicole:
Can the execution time be printed for all test results and not just test failures? I think the execution time would be helpful in all cases especially in test development before the test reaches pass criteria.

PR #281, opened 2017-04-26
TEST_ASSERT_EQUAL with logical behavior

schulz-m:
Currently `TEST_ASSERT_EQUAL` relays to `TEST_ASSERT_EQUAL_INT` and casts the variables to compare to `INT`. This can lead to unexpected behavior whereas it should really just be the opposite of `TEST_ASSERT_NOT_EQUAL`

### DIFF
--- a/src/unity.h
+++ b/src/unity.h
@@ -96,7 +96,7 @@ void tearDown(void);
 #define TEST_ASSERT_EQUAL_INT16(expected, actual)                                                  UNITY_TEST_ASSERT_EQUAL_INT16((expected), (actual), __LINE__, NULL)
 #define TEST_ASSERT_EQUAL_INT32(expected, actual)                                                  UNITY_TEST_ASSERT_EQUAL_INT32((expected), (actual), __LINE__, NULL)
 #define TEST_ASSERT_EQUAL_INT64(expected, actual)                                                  UNITY_TEST_ASSERT_EQUAL_INT64((expected), (actual), __LINE__, NULL)
-#define TEST_ASSERT_EQUAL(expected, actual)                                                        UNITY_TEST_ASSERT_EQUAL_INT((expected), (actual), __LINE__, NULL)
+#define TEST_ASSERT_EQUAL(expected, actual)                                                        UNITY_TEST_ASSERT(((expected) ==  (actual)), __LINE__, " Expected Equal")
 #define TEST_ASSERT_NOT_EQUAL(expected, actual)                                                    UNITY_TEST_ASSERT(((expected) !=  (actual)), __LINE__, " Expected Not-Equal")
 #define TEST_ASSERT_EQUAL_UINT(expected, actual)                                                   UNITY_TEST_ASSERT_EQUAL_UINT( (expected), (actual), __LINE__, NULL)
 #define TEST_ASSERT_EQUAL_UINT8(expected, actual)                                                  UNITY_TEST_ASSERT_EQUAL_UINT8( (expected), (actual), __LINE__, NULL)


schulz-m:
I see that the CI is failing, someone more experienced please tell me if this change makes sense and then I could have a look into it :-)

Letme:
It kinda breaks backwards compatibility - even if you are right, but lets consider that integer is a basic value so internally it does same as your macro except that it uses some casting.

I must say kudos to unit tests - they caught exactly that:
```
../src/unity.h:99:130: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
```

schulz-m:
What do you mean with "internally it does the same"? In our use case we also use the Unity framework to test C++ code (as we have both C & C++ in our embedded systems), and the cast doesn't work well with classes...

Letme:
Internally `UNITY_TEST_ASSERT_EQUAL_INT` calls function ```UnityAssertEqualNumber``` which beside casting does the (negative) check.

So, can you give short example where the cast does not work well with classes?

schulz-m:
Mmh excuse me - I somehow have trouble understanding your point. It doesn't work with any class that doesn't have a valid cast to `UNITY_INT`. In our example we have a `Matrix` class and we can use `TEST_ASSERT_NOT_EQUAL` but not `TEST_ASSERT_EQUAL` because of it doesn't just use boolean comparison. Obviously this is not a big deal as we can just use the comparison within `TEST_ASSERT_TRUE` but I still found it ugly, hence the proposed change.

mvandervoord:
The intention of TEST_ASSERT_EQUAL isn't to compare any types. That can't
be done reliably in a statically typed language like C. It is an alias for
comparing integer types, as these are the most common comparisons.

schulz-m:
I see, I was under a different impression as I looked at `TEST_ASSERT_NOT_EQUAL` - but then feel free to reject the PR :-)

mvandervoord:
@schulz-m I definitely understand your point. When a new user comes across TEST_ASSERT_EQUAL, they are likely going to assume that they can throw any two objects of the same type in there and they will be compared. 

Because of this, I am tempted to merge this kind of change... but I have a few things keeping me from doing so:

(1) It breaks backwards compatibility. The TEST_ASSERT_EQUAL macro was one of the first macros to be included with Unity and many people have been using it as an alias for int. Their code should still work, most of the time (with some possible casting issues here and there) but they'd miss out on detailed reporting. 

(2) There are going to be types that it just can't compare and people are going to wonder why it broke. Stick two structs in there and you're going to get a cryptic compiler error. This is a limitation of the language, unfortunately.

The asymmetry of TEST_ASSERT_EQUAL and TEST_ASSERT_NOT_EQUAL is definitely weird. 

So I see a few options for handling this. If anyone has opinions, I'd love to hear them:

A - Make TEST_ASSERT_EQUAL compare whatever you give it. If it's not legal in the language, then they'll get a compiler error and that's what documentation is for. If they wanted more detail to be reported, they should be using a more specific assertion.

B - Make TEST_ASSERT_NOT_EQUAL an alias for int just like EQUAL. If people get the occasional unexpected casting false-pass, that's what documentation is for. ;)

C - Drop these two assertions altogether. People will just have to update their code when they update to this version. 

Thoughts?

Letme:
I would pick C - since we want people to just be explicit all the time, so that they know what they are doing.

mvandervoord:
I have posted a poll to solicit opinions about this one. I'd appreciate it if you took a moment to vote. Thanks! http://www.throwtheswitch.org/articles/2017/11/1/please-vote

mvandervoord:
This issue has been resolved by adding additional options.

Issue #274, opened 2017-03-27
New test suggestion: File Compare


dpostorivo:
I've been testing a lot of file IO recently and one thing that would be a nice to have would be a file compare inside of unity. For now I've been opening both files, putting them in a buffer, then comparing the resulting buffers and their size. 

Obviously this would not be enabled by default especially in most embedded contexts, but having it will simplify tests for those who do work with file IO.

Thoughts on this? I can get started if others are interested.

mvandervoord:
I think that's a really good idea. I agree that it would be something that is disabled by default. We have other things that fall into this category already, though. I think it makes a lot of sense.

Personally, I can imagine two useful versions:

TEST_ASSERT_EQUAL_TEXT_FILE - loads the files and then uses TEST_ASSERT_EQUAL_STRING_ARRAY or something?
TEST_ASSERT_EQUAL_BINARY_FILE - loads the files and then uses TEST_ASSERT_EQUAL_MEMORY

:)

dpostorivo:
So I've tried to start making the binary file first which is in #283 . If someone could review it that would be nice.

AndersonMartins1:
To resolve the problem described, you can follow the steps below:

Add unity_config.h to the Unity Include Directory:

Make sure the directory containing unity_config.h is added to Unity's include directories during compilation. This can be done by specifying -I<path_to_your_directory> in the build options.
Configure UNITY_INCLUDE_CONFIG_H:

Define the UNITY_INCLUDE_CONFIG_H macro during compilation to enable the required doubles. This can be done using -DUNITY_INCLUDE_CONFIG_H in the build options.
Update CMakeLists.txt:

In the CMakeLists.txt file in the testing directory, add the directory containing unity_config.h to the target unity include directories. This can be done using target_include_directories(unity PUBLIC <path_to_your_directory>).
Check Directory Hierarchy:

Verify that the directory hierarchy is correct and that unity_config.h is located in a location accessible to Unity during compilation.
Test the Configuration:

After making the above changes, compile your project with CMake and check that Unity recognizes unity_config.h correctly and that doubles are enabled as expected.
Additional Debugging:

If the problem persists, check that there are no typos or incorrect paths in the CMakeLists.txt settings and that there are no conflicts with other compilation settings or flags.
Make sure you follow these steps carefully and check that the settings are correct at each step. If the problem persists, carefully review the error messages and try to identify the root cause of the problem to effectively resolve it.

[...]

PR #283, opened 2017-05-07
Added binary file compare asserts

dpostorivo:
First part of #274.  Since this is the first I've contributed to the unity core I would just like to know this is the right direction to go for this.


mgiaco:
Cool, are there any plans to add this to the mainline?

.gitignore
```diff
@@ -1,9 +1,15 @@
 build/
 test/sandbox
 .DS_Store
+
```
mgiaco: I think it is not good to add new things to the ignore file

src/unity.c
```diff
@@ -42,8 +42,11 @@ static const char UnityStrNaN[]                    = "NaN";
 static const char UnityStrDet[]                    = "Determinate";
 static const char UnityStrInvalidFloatTrait[]      = "Invalid Float Trait";
 #endif
+#ifndef UNITY_EXCLUDE_FILE
```
mgiaco: I think thats wrong here?

```C

#ifndef UNITY_EXCLUDE_FILE
const char UnityStrErrFile[]                       = "Unity File Comparison Disabled";
#endif
```

src/unity.c
```diff
@@ -1153,6 +1156,54 @@ void UnityAssertEqualMemory(UNITY_INTERNAL_PTR expected,
...
+
+    FILE* expectedFile = NULL;
+    FILE* actualFile = NULL;
+    UNITY_UINT32 expectedSz, actualSz;
+    UNITY_PTR_ATTRIBUTE unsigned char* ptr_exp = NULL;
+    UNITY_PTR_ATTRIBUTE unsigned char* ptr_act = NULL;
+
+    RETURN_IF_FAIL_OR_IGNORE;
+
+    expectedFile = fopen(expected, "r");
```
mgiaco: I had some problems here after i changed it to "rb" everything works fine.

src/unity.c
```diff
@@ -1153,6 +1156,54 @@ void UnityAssertEqualMemory(UNITY_INTERNAL_PTR expected,
...
+    if (expectedFile == NULL)
+        UNITY_FAIL_AND_BAIL;
+    fseek(expectedFile, 0, SEEK_END);
+    expectedSz = (UNITY_UINT32) ftell(expectedFile);
+    fseek(expectedFile, 0, SEEK_SET);
+    ptr_exp = malloc(expectedSz);
+    fread(ptr_exp, 1, expectedSz, expectedFile);
+    fclose(expectedFile);
+
+    actualFile = fopen(actual, "r");
```
mgiaco: Same as above i also changed this to "rb" because of some memory compare problems.

src/unity.h
```diff
@@ -205,6 +205,12 @@ void tearDown(void);
 #define TEST_ASSERT_DOUBLE_IS_NOT_NAN(actual)                                                      UNITY_TEST_ASSERT_DOUBLE_IS_NOT_NAN((actual), __LINE__, NULL)
 #define TEST_ASSERT_DOUBLE_IS_NOT_DETERMINATE(actual)                                              UNITY_TEST_ASSERT_DOUBLE_IS_NOT_DETERMINATE((actual), __LINE__, NULL)
 
+/* File (If Enabled) */
```
mgiaco: Here the indention is not correct - code style

src/unity.h
```diff
@@ -205,6 +205,12 @@ void tearDown(void);
...
+#define TEST_ASSERT_BINARY_FILE(expected, actual)                                         UNITY_TEST_ASSERT_BINARY_FILE((expected), (actual), __LINE__, NULL)
```
mgiaco: Here the indention is not correct - code style

src/unity.h
```diff
@@ -332,6 +338,13 @@ void tearDown(void);
 #define TEST_ASSERT_DOUBLE_IS_NOT_NAN_MESSAGE(actual, message)                                     UNITY_TEST_ASSERT_DOUBLE_IS_NOT_NAN((actual), __LINE__, (message))
 #define TEST_ASSERT_DOUBLE_IS_NOT_DETERMINATE_MESSAGE(actual, message)                             UNITY_TEST_ASSERT_DOUBLE_IS_NOT_DETERMINATE((actual), __LINE__, (message))
 
+/* File (If Enabled) */
```
mgiaco: Here the indention is not correct - code style

src/unity_internals.h
```diff
@@ -780,5 +796,15 @@ int UnityTestMatches(void);
 #define UNITY_TEST_ASSERT_DOUBLE_IS_NOT_DETERMINATE(actual, line, message)                       UnityAssertDoubleSpecial((UNITY_DOUBLE)(actual), (message), (UNITY_LINE_TYPE)(line), UNITY_FLOAT_IS_NOT_DET)
 #endif
 
+#ifdef UNITY_EXCLUDE_FILE
```
mgiaco: Here the indention is not correct - code style

test/support/example.bin
```diff
@@ -0,0 +1 @@
+	
```
mgiaco: i think it would be better to use the  "test\expectdata" folder for that - so why adding a new folder?

mvandervoord:
I think some of the logic may need to be reversed. It's more common for embedded developers to NOT have access to file support, so I would say that the default should be to not include the file support and to have to purposefully enable it.

dpostorivo:
@mgiaco  I will look into the comments the next few days thanks. 

@mvandervoord Let me know the specific logic an I can work on it. I was trying to base what I was doing on float support, but if that isn't the path I need to take let me know.

mgiaco:
Any news here?

dpostorivo:
Sorry, I've been pretty busy at work recently. I'd like to try and pick it
up again soon.

.gitignore
dpostorivo: I can drop this, but swap files are pretty common on ignores.

mgiaco:
Hey no problem many thanks.

src/unity.c
dpostorivo: Since it's using memory compare to actually run the test, I'm not quite sure if the message will actually be used. I'll put in in there so if someone has a better idea of how it should be structured, it will be available.

mgiaco:
Any news here? What i also found out is that the size is not checked. So that could be a problem. As a workaround i have added some test after the memory check.

```C
    UnityAssertEqualMemory(ptr_exp, ptr_act, expectedSz, 1, msg, lineNumber, flags);

    TEST_ASSERT_EQUAL_UINT32(expectedSz, actualSz);
```

mvandervoord:
Sigh. I still like this concept... but we're tracking a moving target. 

Would it simplify things if comparing files became an extra header in the extras folder? So a user could just pull in `"unity_files.h"` to get the extra file assertions?

dpostorivo:
Maybe, but the merge confilcts are the least of my worries on this.


 I'll be honest when I started, I didn't have a good hold of what the internals of unity looked like. Now I have a better grasp (no perfect), I'd like to reevaluate what I've done in the past.  

It may also make sense to start from scratch.

mvandervoord:
That makes sense to me. How about this, then? I'm going to close this PR. I'd definitely welcome another attempt when / if you have time to do so. I think the concepts is great and would appreciate it's addition to the project. I DO think there are enough systems where file-systems are not part of the system that it makes sense to have it as a separate include... so maybe aim that way. In the long run, we might end up with multiple asserts... simples ones that do binary / string assertions... and eventually maybe smarter ones that do more complex comparisons between more structured data? (certainly don't wait on that later thing for a first PR)

mgiaco:
@dpostorivo  do you still work on that?

dpostorivo:
@mgiaco not right now, and I probably won't have time until a few weeks into November. It's still something I would like to see in Unity, but I don't quite have the time.

PR #286, opened 2017-08-01
Allow specifying custom header declaration

palaviv:
Currently when you define custom `UNITY_OUTPUT_FLUSH` without `UNITY_OMIT_OUTPUT_FLUSH_HEADER_DECLARATION` you add a declaration of the flush function. That causes a compilation problem.
For example:
```c
#define UNITY_OUTPUT_FLUSH()    RS232_flush()
```
Then we get the following declaration:
```c
extern void UNITY_OUTPUT_FLUSH(void)
```
That result is the following error:
```
error: macro "UNITY_OUTPUT_FLUSH" passed 1 arguments, but takes just 0
 extern void UNITY_OUTPUT_FLUSH(void);
```

The `UNITY_OUTPUT_CHAR` declaration actually works because when you define:
```c
#define UNITY_OUTPUT_CHAR(a)    RS232_putc(a)
```
The declaration works because:
```c
extern void UNITY_OUTPUT_CHAR(int)
```
is being translated to:
```c
extern void RS232_putc(int)
```

We can probably solve that problem with a patch by changing:
```c
extern void UNITY_OUTPUT_FLUSH(void);
```
to:
```c
extern void UNITY_OUTPUT_FLUSH();
```

I suggest making a bigger change that will allow more flexibility by changing `UNITY_OMIT_OUTPUT_CHAR_HEADER_DECLARATION` and `UNITY_OMIT_OUTPUT_FLUSH_HEADER_DECLARATION` to `UNITY_OUTPUT_CHAR_HEADER_DECLARATION` and `UNITY_OMIT_OUTPUT_FLUSH_HEADER_DECLARATION` thus allowing any function declaration.

### DIFF
--- a/examples/unity_config.h
+++ b/examples/unity_config.h
@@ -201,10 +201,12 @@
  * `stdout` option. You decide to route your test result output to a custom
  * serial `RS232_putc()` function you wrote like thus:
  */
-/* #define UNITY_OUTPUT_CHAR(a)    RS232_putc(a) */
-/* #define UNITY_OUTPUT_FLUSH()    RS232_flush() */
-/* #define UNITY_OUTPUT_START()    RS232_config(115200,1,8,0) */
-/* #define UNITY_OUTPUT_COMPLETE() RS232_close() */
+/* #define UNITY_OUTPUT_CHAR(a)                    RS232_putc(a) */
+/* #define UNITY_OUTPUT_CHAR_HEADER_DECLARATION    RS232_putc(int) */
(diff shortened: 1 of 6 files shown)


mvandervoord:
I like your proposed changes.

I am unclear on what the problem is that you are trying to solve, though. Under what circumstance is the flush macro passed an argument?

palaviv:
> I am unclear on what the problem is that you are trying to solve, though. Under what circumstance is the flush macro passed an argument?

Actually the default is `fflush(stdout)` that receive stdout as argument.

PR #287, opened 2017-08-05
Fix custom UNITY_OUTPUT_FLUSH and add tests

palaviv:
This fix a error in the declaration of UNITY_OUTPUT_FLUSH_HEADER_DECLARATION from #286. In addition I added some tests.

### DIFF
--- a/src/unity_internals.h
+++ b/src/unity_internals.h
@@ -263,7 +263,7 @@ extern void UNITY_OUTPUT_CHAR_HEADER_DECLARATION;
 #else
 /* We've defined flush as something else, so make sure we declare it here so it's ready for use */
 #ifdef UNITY_OUTPUT_FLUSH_HEADER_DECLARATION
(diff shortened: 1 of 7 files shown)


palaviv:
Can anyone help me understand what is failing in the CI?

mvandervoord:
It appears that you are failing the parameterized self-tests... I suspect that it might be another instance of needing to define the new flush statement (though I haven't pulled this down and tried it myself yet). 

Sorry about the long delay here. I really appreciate that you put the time into this, particularly since you made an effort to add tests for something we missed! If you get stuck trying to hunt down the problem, I'll help you work it out. Unfortunately since I took so long to get back to you, you now have a conflict on your makefile as well. Again, I apologize!

farrrb:
Fixed with: https://github.com/ThrowTheSwitch/Unity/pull/324

PR #292, opened 2017-08-30
Allow test to assert that expectations so far have been met

tbates-redarc:
For debugging a failing test with a lot of expects, I frequently would like to be able to assert that all the expects I have set up so far have been met, rather than waiting until the end of the test. Since usually if an expect is not met I just get the line number of the test, it can be hard to tell how far through the test got before the missing expectation.

This is not a complete pull request, just a demonstration of how my proposed feature could potentially work - looking for input on the best way to implement it, which header file to put it in, what to call it and so on. Please let me know how you would like it done and I will clean it up before you pull.

### DIFF
--- a/auto/generate_test_runner.rb
+++ b/auto/generate_test_runner.rb
@@ -217,7 +217,7 @@ class UnityTestRunnerGenerator
     end
     output.puts("}\n")
 
-    output.puts('static void CMock_Verify(void)')
+    output.puts('void CMock_Verify(void)')
     output.puts('{')
     mocks.each do |mock|
       mock_clean = TypeSanitizer.sanitize_c_identifier(mock)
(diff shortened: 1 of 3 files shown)


mvandervoord:
Feel free to reopen this if I have misunderstood, but I think what you are looking for is already generated. In the test runner, it creates a function that you can use in your tests called resetTest(). If you call it, it verifies all your mocked calls thus far and resets everything for the next batch.

tbates-redarc:
It's similar, but I just want to call `CMock_Verify()` without also calling `CMock_Destroy()`, `tearDown()`, `CMock_Init()` and `setUp()`. I want to assert that all of the expectations set up so far have been met by the calls so far, and then go on to do other things.

For example I have a state machine that requires multiple calls to its `update()` function to get where it needs to go. I write tests like:

```
void test_state_machine(void)
{
  things_happen_Expect();
  kick_off_process();

  more_things_happen_Expect();
  update();

  yet_more_things_happen_Expect();
  update();
}
```

I want to be able to assert that `more_things_happen()` is called after the first `update()` and not the second one. As it is today, this test will pass if the first `update()` does nothing but the second `update()` calls both `more_things_happen()` and `yet_more_things_happen()`.

PR #296, opened 2017-09-13
Allow suiteSetUp() and suiteTearDown() to be provided as normal C functions

jlindgren90:
This is simpler and more flexible than embedding C code in the Ruby options
(:suite_setup and :suite_teardown).  However, support for :suite_setup and
:suite_teardown is kept for backwards compatibility.

Several configurations are possible:
1. :suite_setup and :suite_teardown options provided and used.
2. :suite_setup and :suite_teardown options not provided (nil):
  2a. Weak symbols not supported; suiteSetUp() and suiteTearDown() are not called.
      It would be simpler to make user-provided functions mandatory in this case,
      but it could break some pre-existing test suites.
  2b. Weak symbols are supported and the stub implementations of suiteSetUp() and
      suiteTearDown() are called if there are no user-provided functions.
  2c. Weak symbols are supported but overridden by user-provided suiteSetUp() and
      suiteTearDown() functions.


mvandervoord:
Heya. I was about to merge this, but then I noticed that you've split unity_internals into an extra file. We've purposefully kept this project at the three main files and would like to continue to do so. If you move the weak definitions back to internals, I'll happily merge everything. :)

jlindgren90:
Done.

PR #297, opened 2017-09-14
Print numbers in higher precision in UnityPrintFloat

jsalling:
The new implementation of `UnityPrintFloat` was mostly worse than the existing one. This branch reverts the changes and then adds back the key features of that branch.

The new attempt had a few drawbacks. It only printed 6 digit of precision, whereas the existing code printed 9+ correctly. You need 9 digits to differentiate single-precision floating point numbers. It had mistakes, rounding errors in the 6th digit on millions of float values. Finally, it covered up the correctness problems by using tests that ignored the last digit.

The existing implementation didn't attempt to print very small values (< 0.000001). Don't throw the baby out with the bathwater! It also didn't print negative zero as `-0.0`.
Let's keep these two ideas: printing `-0.0`, and print 6 digits (for now) for very small float values.
* Revert "Merge pull request #291"
* Cherry pick code attributed to author of commit 0e7eb54
* Clean up and match printing style


mvandervoord:
Hi. I'll definitely check through this. I'd like our floating point representation to be as accurate as possible.

Your opening premise has a flaw, though. Single precision floating point is accurate to 6 decimal points and (7 occasionally). It stores the decimal portion in 23 bits, making our maximum perfectly represented number 8,388,608. Everything larger than that has the potential to be missing information.

FSMaxB:
It's 24 bits. IEE754 floating point numbers don't store the leading zero as part of the mantissa. And full roundtrip conversion might require 9 digits.

mvandervoord:
@FSMaxB, I believe you are accidentally counting the sign bit. It's 23 digits of actual numerical range. Even if I were wrong about this, though, that only gets you 2^24 which would be 7 significant digits (with an occasion of 8, when the highest significant bit is 1).

The leading zero isn't a significant digit, so it's not really part of the conversation. Significant digits are those that you can fact check for accuracy. For example, if you chose to print without exponents, both of these could be guaranteed to print correctly:

0.1234567
0.000000000000001234567

While we are showing many more digits in the second example, it still only has 7 significant digits (those that are not leading / trailing zeros). 

These would NOT be guaranteed to print correctly

0.87654321
0.0000000000000087654321

Even the first one would be a problem because it has 8 significant digits, meaning that it would have to store 87654321 as a 23 (or even 24 if you count the sign) bit int.

If I'm explaining this poorly, here are some references which are decent:
https://stackoverflow.com/questions/1883385/rounding-to-use-for-int-float-int-round-trip-conversion
https://en.wikipedia.org/wiki/IEEE_754-1985 (I include this one, because it has a nice table that shows the precision)
https://en.wikipedia.org/wiki/IEEE_754 (this one is the newer standard and gets into much more detail)

FSMaxB:
No, I'm not counting the sign bit. A mantissa of 23 zeroes is actually a one followed by 23 zeroes, since an actual zero is represented in the exponent, the mantissa is encoded in such a way that the leading `1` is omitted.

Also Wikipedia states that single precision floating point number represent values with 6 to 9 significant decimal digits. https://en.wikipedia.org/wiki/Single-precision_floating-point_format#IEEE_754_single-precision_binary_floating-point_format:_binary32

From what I understand this doesn't mean that you can losslessly represent 9 decimal digits in the binary representation, but it may be necessary to use 9 digits in order to be able to parse it back to the same binary represenation (but not entirely sure about that).

mvandervoord:
Ah yes. I agree with you. I see we are talking about opposite sides of this squishy subject. 

I was arguing that if you convert to a string, you can only guarantee that 6 digits are correct (sometimes more will be correct, but sometimes they won't). 

You are arguing that if you want to convert back from a string to a number, that 9 digits are required to make sure it's the full number (some of these digits will be unnecessary for most numbers, but you must have 9 for the worst case numbers).

Both of these are correct. Sorry that I was misunderstanding your point.

When displaying test results, I felt that we should show only the precision that we are confident is correct because I wouldn't want to say a number was 0.12345678 when really it was 0.12345677... I believe you are saying (and please correct me if I am wrong) that we should display all digits we can because sometimes there might actually be information in that 8th or 9th digit that WAS the failure. Yes?

I'm torn honestly. I hate to print misleading information, which can happen in either case (it's just misleading in a different direction).

jlindgren90:
It's true that 6 digits isn't sufficient to do a bit-accurate round-trip from decimal back to floating-point.  If that's the goal, then you need more than 6 digits in many cases.  It's also true that if you try to print 9 significant digits using purely single-precision math, you're guaranteed to be printing incorrect digits in many cases.  Pick your poison.  One approach is not objectively "worse" than the other; it depends on what you are trying to accomplish.

The gold standard for printing floating point values is the "Dragon 4" algorithm, which to my understanding uses integer math in larger-than-machine-word precision.  It claims to be able to print the minimum number of digits necessary to round-trip back to float (so for example, it would print "0.1" and not "0.100000001", since they both convert to the same float) while also printing exactly correct digits.  In my mind that's overkill and too much complexity for an embedded system, though.  The whole point of having a custom UnityPrintFloat routine was to have something simpler than the libc version, and that simplicity comes with an accuracy trade-off.

FSMaxB:
I'm not quite sure about which one of both options is better. I just wanted to explain where the 9 digits in @jsalling's comment came from.

FSMaxB:
@jlindgren90 There's also Grisu3 since 2010.

mvandervoord:
:) Thanks for the good conversation, both of you! I'm in agreement that we're staying "simple" for a reason and we're just going to have to eat some inaccuracies. My personal feeling has been that it's better to show less digits (though it does create the occasional curiosity like "Expected 0.876543 Was 0.876543" because the problem was in the 7th+ decimal place). In those cases, I just hope people understand that floating point printing is tricky and we're not aiming to be the gold standard. 

I suppose we could always give people the option to use the one built into libc if they want, or work with ours otherwise.

I ANY CASE, this has been a good conversation and my plan is still to look through your stuff and likely submit it. I really appreciate you people! Thanks!

src/unity.c
```diff
@@ -235,97 +235,121 @@ void UnityPrintMask(const UNITY_UINT mask, const UNITY_UINT number)
...
-        /* scale up or down by powers of 10 */
+        /* scale up by powers of 10 */
         while (number < 100000.0f / 1e6f)  { number *= 1e6f; exponent -= 6; }
         while (number < 100000.0f)         { number *= 10.0f; exponent--; }
-        while (number > 1000000.0f * 1e6f) { number /= 1e6f; exponent += 6; }
-        while (number > 1000000.0f)        { number /= 10.0f; exponent++; }
 
         /* round to nearest integer */
-        n = ((UNITY_INT32)(number + number) + 1) / 2;
+        n = (UNITY_INT32)(number + 0.5f);
```
jlindgren90: Doing the rounding in integer math was intentional, because an int32 has more precision than a float.  If you do it in float, you can completely lose the 0.5f due to truncation and get incorrect rounding in some cases.

src/unity.c
```diff
@@ -235,97 +235,121 @@ void UnityPrintMask(const UNITY_UINT mask, const UNITY_UINT number)
...
+        UNITY_OUTPUT_CHAR('-');
+        if (exponent < 10) UNITY_OUTPUT_CHAR('0');
+        UnityPrintNumber(exponent);
+    }
+    else if (number < 4294967295.9999995) /* Rounded result fits in 32 bits, "%.6f" format */
+    {
+        const UNITY_INT32 divisor = 1000000/10;
+        UNITY_UINT32 integer_part = (UNITY_UINT32)number;
+        UNITY_INT32 fraction_part = (UNITY_INT32)((number - (UNITY_DOUBLE)integer_part)*1000000.0 + 0.5);
+        /* Double precision calculation gives best performance for six rounded decimal places */
```
jlindgren90: Can we/do we want to rely on having double-precision support?

mvandervoord: Definitely no. A lot of the supported targets don't have double precision... and we're unlikely to ever print that many significant digits anyway. ;)

jlindgren90:
I also had a working version that would print 7 significant digits, which with a few more lines of code (mainly doing the powers-of-10 scaling more carefully) was correct in 80% of cases but could be off-by-one in the last digit for some cases.  I can submit that as a new pull request if there is interest in a "compromise" solution.

mvandervoord:
@jlindgren90 That would probably be a good idea.

src/unity.c
jsalling: That's only true outside the range of numbers we are dealing with here. `100,000` to `1,000,000` can always add `0.5f` within the precision. But you knew that ( I can tell you know a thing or two about floats).
I thought this code was cleaner, showing more intent. It's a judgement call.

src/unity.c
jsalling: It's another trade-off, as folks mentioned above. You need more bits of precision to accurately print more digits. With a 23 bit mantissa the calculation cannot even round the 6th digit correctly a lot of the time. Surprising, huh.
You can eliminate the use of doubles by excluding the code. `UNITY_EXCLUDE_FLOAT_PRINT` handles that option for tight embedded systems.

jsalling:
I love that this sparks such a good, meaningful conversation. It's amazing to me that within half a day so many contributors are discussing intricacies of floats. Wow.
I guess you know where I stand on the choices here. I like getting 9-16 digits of precision printed...  correct to the last digit, rounded right, and with options to round-to-even or round-away-from-zero.

This feels like a good compromise to me, but I totally see the appeal of simpler code, lower-bit math, and an embedded focus. That trade would necessarily print fewer digits.

Cheers.

src/unity.c
jlindgren90: Hmm, yes, this mattered for the version I had locally that was printing 7 digits but doesn't matter for 6.  Good point.

mvandervoord:
Hey guys. I'm hoping I can respark some conversation here, despite my long lag in communication. After giving this a lot of thought, I have a proposal and I'd like to hear what you all have to say about it.

My proposal is this. We start by merging Pull Request #299. This is the "compromise" solution that shows 7 digit float results, which are usually correct for single-precision and always correct for double-precision. The algorithm is fairly compact and should work on most embedded processors where software floating point is used (or even smaller processors with hardware floating point). 

A large part of our population consists of embedded software developers (myself included) and I want to make sure Unity works for them out of the box (or as close to that as we can get with a quirky language like C!).

THEN, I'd like to add another option (something like UNITY_INCLUDE_FLOAT_PRINTF) which uses C standard library print functions to print them. Those algorithms cover all the weird cases we're worried about, so there's not much point in us trying to develop our own.

This gives us three levels:

- No Floating Point Printing (for tiny processors)
- Simple Floating Point Printing that is mostly correct (for many ARM processors, etc)
- Full Floating Point Printing that is always correct (for native tests, larger footprint processors, etc).

I *think* this makes everyone happy, without increasing our complexity too much.

What do you guys think? ( and feel free to tell me I'm an idiot. I have thick skin )    :)

jlindgren90:
Sounds good to me.

bjones1:
+1 from me as well.

Letme:
+1 form me. Maybe adding that 2nd option would be default and all of it nicely described in docs?

mvandervoord:
Good points, @Letme. I was already on the same page with the 2nd option being the default. Good documentation is always a good thing to remember!

jsalling:
Mark @mvandervoord , you know I have to disagree. I can't help it. ;)

I like most of the proposal. I would change it to merge this PR as the starting point. #297 
I support using the C standard library as an option. It makes sense to follow up with that. That way `double` values could be even better.

I'm going to make one more attempt to sum up the technical merits of this PR. First, it prints 9 digits. Established earlier in this thread, this is the number of digits needed to differentiate between two successive binary `float` numbers, when those numbers are printed in base-10 for human consumption. Error messages are for humans (usually) and making numerical error messages clearer is good. 
To dispel any misconceptions - Objectively, the pros of #297 :
* Prints float values correctly
* Prints enough digits, 9, to round-trip to the offending `float` number
* Readable and compact code
* Runs on ARM embedded - on a Cortex-M processor with software float or A8 hardware, and others
* Works for embedded in general - except where no software float or double is available
* Rounds ties to even, this is the default rounding behavior for most standard libraries
* Option to round ties away from zero
* Adds -0 and very small number support
* Prints 42.0 (with a trailing 0) to imply the underlying type - as in Ruby, Python, etc.


The other PR has improved from where it started. The spirit is in the right place. It now uses some numerical optimization from the existing implementation to improve accuracy.
Comparison, from #299 :
* Prints float values - but many errors in the last digits
* Prints 7 digits - not enough to identify the exact offending `float`
* Code is longer, and complexity is higher by any metric your tools use
* Doesn't follow Unity's coding standard - this is easily fixed
* Works for embedded in general - except where no software float is available
* Rounding to even is not addressed
* No options to set rounding
* Adds -0 and small number support
* Prints 42 (without a 0) - closer to an integer format

That's the technical pro's and con's.
If what is really desired is just 7 digits, I honestly think it would be better to start here and just modify the implementation to print a few less digits.
For a technical subject like this, the more technical solution usually makes sense. In this case, its code is simpler, too. I just keep asking myself, why should we prefer the solution that gives up on the technical merits?

jlindgren90:
I think we are familiar with the pros and cons of each solution by now.  Claiming that one solution "gives up on the technical merits" is a personal opinion, could be taken as mildly insulting, and doesn't really add to the discussion.

I think the choice comes down to two questions:
 - Do we want to rely on having double-precision?  If yes, then I agree, my code is overly complex and you can use something simpler since you don't have to be as careful about precision loss.
 - Do we want to print only the digits that we are reasonably certain about (7) or print enough digits to round-trip the float in all cases (9), even if the last couple of digits are garbage?  You can make a reasonable argument for either approach; one isn't objectively "better" or "clearer" than the other.

Smaller differences, like whether we print "42" or "42.0", could easily be adjusted (in either implementation).

mvandervoord:
One thing I can say for sure is that we can't rely on having double precision. I would like to SUPPORT having double precision, but we can't rely on it. I know that at least some of the floating point targets we support can only handle single precision and won't even fake it in software floating point libraries. So that's a hard line on that one.

@jsalling is the reliance on doubles essential to this solution or was that just a way to simplify a detail?

As for the number of digits to print, I agree @jlindgren90, it's definitely going to come down to personal opinion. I definitely see the argument for printing 9 digits because it will ALWAYS have the ability to show the error. It has the downside of printing values that don't exist sometimes. 

Do either of these foundations provide a simpler means of having an option for UNITY_FLOAT_NUM_SIG_DIGITS to be set to 6-9?  ;) If it's easily done, maybe we just provide an option. (queue next discussion for which is the default. ha!)

jlindgren90:
@jsalling You said this version "Prints enough digits, 9, to round-trip to the offending float number" but it looks like it only prints 6 digits for numbers smaller than 0.1.  Is that intentional?

jsalling:
@mvandervoord The reliance on doubles is not essential. The code could be changed to float only and  retain round-trip properties, but would require modification and allowing errors in 9th digit.

@jlindgren90 Please don't get defensive. I tried to show an objective comparison. Your code is great and well reasoned.

Smaller differences are easily worked out, but are worth mentioning because they become the default. I suppose I'm the only one who cares about rounding behavior.

I was a bit blind to the hard line on doubles - gcc, clang, IAR all support doubles for embedded targets. I'd say, this PR is worse on the technical there.
Either one could be reworked to set a variable number of digits. Some of the code gets trickier to be able to do that well.

I just wanted to be clear about some things that seemed murky.

jsalling:
@jlindgren90 
This whole point of this PR from the beginning was that it uses your code for printing small float values to 6 digits. Yes, it's a hybrid of existing code and your code. I thought this would be familiar.

I appreciated the contribution and liked it. I just preferred to keep the 9 digits for bigger numbers, since it was there already.

jlindgren90:
@jsalling It seems weird to have a procedure that prints 9 significant digits for large numbers but only 6 for small numbers.  Can you make it print 9 in all cases?

I also noticed that this version prints "1.0e-02" rather than "0.01".  My personal preference would be for "0.01", since it takes me longer to read scientific notation.  glibc seems set the cutoff for scientific notation somewhere around 0.0001, and I tried to do similar in my implementation.  Would you be opposed to changing this one to do the same?

jsalling:
> Can you make it print 9 in all cases?
Sure. Seems reasonable to me.

> glibc seems set the cutoff for scientific notation somewhere around 0.0001, and I tried to do similar in my implementation. Would you be opposed to changing this one to do the same?

Not opposed at all. In fact, if you'd like to contribute this I'd happily include it. If not, I'll get around to it but it will be delayed a little.

jlindgren90:
I updated pull request #299 to print 9 digits if UNITY_INCLUDE_DOUBLE is defined, and added round-to-even behavior.  @jsalling, please give it a try and see if it addresses your remaining accuracy concerns.

The updated algorithm now matches glibc's printf("%.9g") exactly for all *single-precision* values that I tested, when we have *double-precision* machine support.  With only single-precision machine operations, we still get errors in the last digit sometimes, but I think that's the nature of the beast--we need machine support for more precision than the value we're trying to print.

mvandervoord:
I've merged #299 at this point. We'll let the feedback roll in and decide where to go from there. Thanks for the great discussion and all the hard work.

PR #299, opened 2017-09-14
Allow UnityPrintFloat() to print a 7th digit.

jlindgren90:
/cc @jsalling

This is a bit of a compromise solution re: the discussion on https://github.com/ThrowTheSwitch/Unity/pull/297.

This version prints 7 significant digits always.  It works reasonably well without requiring the use of double-precision (but will do better if double-precision is available).  The 7th digit is *usually* right but can occasionally be off by one, sometimes even two (in the case of very large or small values).

For example:
UnityPrintFloat(7.999999f) prints "7.999999" if using double but "8" if using float
UnityPrintFloat(9.994f) prints "9.994" if using double but "9.994001" if using float

As I've said before, this is basically impossible to fix without (a) resorting to higher-precision math or (b) reducing the number of digits printed to 6.


jlindgren90:
@mvandervoord Ping, does this look okay to you?

PR #300, opened 2017-09-23
Bugfix greater/less than asserts with unsigned int

### LINKED ISSUE #298: Greater/less than assertions fail in 32 bit mode
```C
TEST_ASSERT_GREATER_THAN_UINT32(0, 4000000000ul);
```
`FAIL: Expected 4000000000 to be greater than 0`

This test fails when Unity is compiled as a 32-bit version. Likewise, if there was `..._UINT64` assertion it would fail in any build.

This is because the feature doesn't account for the type of the passed numbers. It assumes everything is signed. Large `UINT` values end up cast as negative numbers.

Also, there's no GREATER_THAN_OR_EQUAL asserts (`a >= e`). These are needed for completeness of integer operators.

Edit: Follow up to PR #264 #285

jsalling:
Bugfix on the greater/less than asserts. They failed when using `unsigned int` types.
Now check the style parameter before doing a compare.

This feature wasn't fully finished yet.
Added 64-bit specific assert macros, support for "... or equal to" asserts, and more continuous integration builds for the future.

Closes #298


jsalling:
@mvandervoord 
Friendly ping on this PR. Have you had a chance to look at this bugfix?

mvandervoord:
very sorry about the delay. Thanks for being so patient! I'm going to try to tackle this soon!

mvandervoord:
Thanks so much @jsalling on working on this. It's GREAT work and definitely is a welcome improvement for Unity!

PR #304, opened 2017-10-19
Color test results using ANSI escape codes

VLambret:
Help error detection by adding specific colors for test results. This
behavior is activated only when unity if compiled with UNITY_COLOR flag.

-----

This small modification really helped me to detect errors on a serial terminal on Windows for test executed on an embedded target. On my Unix terminal, the colored tests look like this (I modified a test to get an error, dont panic !)

unity_colored_tests_screenshot

If you think this feature is not in Unity philosophy no problem, I would perfectly understand. If you think it's a good idea but need modifications, I will be glad to modify my code until it's perfect for you.

Thanks for your great framework,

Victor

### DIFF
--- a/docs/UnityConfigurationGuide.md
+++ b/docs/UnityConfigurationGuide.md
@@ -337,6 +337,13 @@ things anyway, though... so this option exists for those situations.
 _Example:_
         #define UNITY_EXCLUDE_SETJMP
 
+##### `UNITY_OUTPUT_COLOR`
+
+If you want to add color using ANSI escape codes you can use this define.
+t
+_Example:_
+        #define UNITY_OUTPUT_COLOR
+
 
 
 ## Getting Into The Guts
--- a/src/unity.c
+++ b/src/unity.c
@@ -19,10 +19,17 @@ void UNITY_OUTPUT_CHAR(int);
 
 struct UNITY_STORAGE_T Unity;
 
+#ifdef UNITY_OUTPUT_COLOR
+static const char UnityStrOk[]                     = "\033[42mOK\033[00m";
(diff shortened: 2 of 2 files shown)


jsalling:
I'm not sure this fits into the Unity philosophy, but we need @mvandervoord to make that call.

One thing I noticed, in unity_internals.h:
`static inline` is not available in C89, so this function should be refactored to meet Unity's C89 compatibility.
The most direct replacement would be `UNITY_OUTPUT_CHAR` and/or `UnityPrint` in a MACRO function.

mvandervoord:
I like the idea of this, actually. We have a colorizer script in auto, but it would be nice for those who don't use Ruby to still optionally have this functionality.

I do have a question about your approach though. Maybe I'm missing something, but wouldn't it be simpler to just change the definition of UnityStrPass, UnityStrFail, etc when color is enabled?

Also, a purely personal preference: I'd like the define to be called UNITY_OUTPUT_COLOR instead of UNITY_COLOR... it helps to organize it with the other output-effect options. :)

Thanks so much for the work! I hope I'm not scaring you away by being overly picky!

VLambret:
Not scared at all, quite the opposite actually :)

The patch is modified according to your feedbacks :

- `UNITY_COLOR` renamed into `UNITY_OUTPUT_COLOR`
- Ansi color codes are directly in strings. It means the `UnityPrint` is modified to handle those codes.
- With this new approach the code is C89 compliant. the `test/Makefile` specify standard `c99`, it would be nice to change this to `c89` so contributors have a feedback on their local builds. I can provide a pull request so we can discuss this point further if you think it's useful.

Dont hesitate to suggest other improvements.

PR #311, opened 2017-11-20
Floating point casting

Kochise:
float/double pointer missing casting that throw a warning


Kochise:
OK, looks like in the original configuration, it works, while in mine, I get an error. Forget it, I pull the plug.

mvandervoord:
It appears the problem is with const. You probably could have gotten away with casting to a const UNITY_FLOAT (though some compilers would give warnings about it being redundant. It seems we can't please all compilers!)

Issue #317, opened 2018-01-09
unity.h should include <stddef.h>


zlalanne:
unity.h makes several references to "NULL" which is defined in <stddef.h>. However unity.h does not actually include <stddef.h>.

GCC compiler #include <stddef.h> in several other standard header files (string, malloc, time, stdlib...) so most cases this will already be available. But this isn't true for all compilers, and because there is a dependency on "NULL" being defined, <stddef.h> should be included.

farrrb:
This issue is resolved and can be closed :)

PR #318, opened 2018-01-10
Add <stddef.h> to unity.h

zlalanne:
unity.h uses the NULL value that is defined in <stddef.h>, however it does
not explicitly include <stddef.h>.

Resolves issue #317

### DIFF
--- a/src/unity.h
+++ b/src/unity.h
@@ -13,6 +13,7 @@ extern "C"
 {
 #endif
 
+#include <stddef.h>
 #include "unity_internals.h"
 
 /*-------------------------------------------------------


jeremyhannon:
seems like a good suggestion to explicitly declare `#include` dependencies. I tested it locally with our Microsoft Visual Studios compiler and IAR RX compiler and didn't see any compile issues.

mvandervoord:
You'll need to surround this with #ifndef UNITY_EXCLUDE_STDDEF_H ... #endif so that Unity can still be used on embedded platforms without stddef (which are numerous, sadly).

zlalanne:
@mvandervoord I can update the PR by guarding using UNITY_EXCLUDE_STDDEF_H. Is there a set of places I need to document this #define?

mvandervoord:
Thanks for asking! In the docs folder there is a UnityConfigurationGuide.md. If you could put it there in "The Options" section, that would be really awesome.

mvandervoord:
Replaced by more complete PR #359

PR #319, opened 2018-01-29
Some minor changes for parse_output.rb

farrrb:
If i understand correctly this is the script for the raw unity output to JUnit style xml and overall test summary. I modified the script in four different ways to get it working:

- The String split now works correctly didn't work correct on windows ("\") platforms and unix based ("/") platforms. 

- Removed the unnecessary whitespaces in the xml output, i needed this for the interpretation of the tools. It was really ugly and sometimes there were 2 whitespaces at front of a message. 

- Added support for the TEST_IGNORE() macro which outputs no message. This wasn't considered in the summary at all.

- Added the name property to the testsuite tag in the xml output for completeness reasons.

### DIFF
--- a/auto/parse_output.rb
+++ b/auto/parse_output.rb
@@ -47,7 +47,12 @@ class ParseOutput
 
     @test_flag = true
     # Split the path name
-    test_name = test_suite_name.split('/')
+    test_name = if @class_name == 1
+                  test_suite_name.split('\\') # Windows
+                else
+                  test_suite_name.split('/')  # Unix based
+                end
+
     # Remove the extension
     base_name = test_name[test_name.size - 1].split('.')
     @test_suite = 'test.' + base_name[0]
@@ -78,11 +83,11 @@ class ParseOutput
     @array_list.push '     <testcase classname="' + test_suite + '" name="' + test_name + '"/>'
   end
 
-  # Test was flagged as being ingored so format the output
+  # Test was flagged as being ignored so format the output
   def test_ignored(array)
     last_item = array.length - 1
     test_name = array[last_item - 2]
-    reason = array[last_item].chomp
+    reason = array[last_item].chomp.lstrip
     test_suite_verify(array[@class_name])
     printf "%-40s IGNORED\n", test_name
 
@@ -96,7 +101,7 @@ class ParseOutput
     return unless @xml_out
 
     @array_list.push '     <testcase classname="' + @test_suite + '" name="' + test_name + '">'
-    @array_list.push '            <skipped type="TEST IGNORED"> ' + reason + ' </skipped>'
+    @array_list.push '            <skipped type="TEST IGNORED">' + reason + '</skipped>'
     @array_list.push '     </testcase>'
   end
 
@@ -104,7 +109,7 @@ class ParseOutput
   def test_failed(array)
     last_item = array.length - 1
     test_name = array[last_item - 2]
-    reason = array[last_item].chomp + ' at line: ' + array[last_item - 3]
+    reason = array[last_item].chomp.lstrip + ' at line: ' + array[last_item - 3]
     test_suite_verify(array[@class_name])
     printf "%-40s FAILED\n", test_name
 
@@ -118,11 +123,11 @@ class ParseOutput
     return unless @xml_out
 
     @array_list.push '     <testcase classname="' + @test_suite + '" name="' + test_name + '">'
-    @array_list.push '            <failure type="ASSERT FAILED"> ' + reason + ' </failure>'
+    @array_list.push '            <failure type="ASSERT FAILED"> ' + reason + '</failure>'
     @array_list.push '     </testcase>'
   end
 
-  # Figure out what OS we are running on.   For now we are assuming if it's not Windows it must
+  # Figure out what OS we are running on. For now we are assuming if it's not Windows it must
   # be Unix based.
   def detect_os
     os = RUBY_PLATFORM.split('-')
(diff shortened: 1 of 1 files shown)


jeremyhannon:
changes look ok to me. We've been using this script in production on Windows Server 2012, converting to JUnit for consumption by Jenkins plugin and haven't had any issues before this PR. Oh, and we're using unity_fixture, which affects which type of format gets converted to JUnit.

farrrb:
I found another leading whitespaces which i removed. We are currently not using `unity_fixture.h` in our environment. My modifications are mostly for beauty reasons. I removed the leading whitespaces in order to get a cleaner output (xml) and i changed the path decomposition according to the current platform. I got it runing before on windows servers so it wasn't "broken" but it was a little bit too ugly ;) 

The only "bug" i fixed is that the TEST_IGNORE() (no message) tests were not counted at all. 

*Before:*
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuite tests="4" failures="2" skips="1">
     <testcase classname="test.\work\_git\Unity\src\test_Jenkins" name="test_function_should_doBlahAndBlah">
            <failure type="ASSERT FAILED">  Expected 3 Was 4 at line: 33 </failure>
     </testcase>
     <testcase classname="test.\work\_git\Unity\src\test_Jenkins" name="test_always_true"/>
     <testcase classname="test.\work\_git\Unity\src\test_Jenkins" name="test_always_false">
            <failure type="ASSERT FAILED">  Expression Evaluated To FALSE at line: 23 </failure>
     </testcase>
     <testcase classname="test.\work\_git\Unity\src\test_Jenkins" name="test_ignore_message">
            <skipped type="TEST IGNORED">  this is stupid, isn't it? </skipped>
     </testcase>
</testsuite>
```

*After:*
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="test.test_Jenkins" tests="5" failures="2" skips="2">
     <testcase classname="test.test_Jenkins" name="test_function_should_doBlahAndBlah">
            <failure type="ASSERT FAILED">Expected 3 Was 4 at line: 33</failure>
     </testcase>
     <testcase classname="test.test_Jenkins" name="test_function_should_doAlsoDoBlah">
            <skipped type="TEST IGNORED">No reason given</skipped>
     </testcase>
     <testcase classname="test.test_Jenkins" name="test_always_true"/>
     <testcase classname="test.test_Jenkins" name="test_always_false">
            <failure type="ASSERT FAILED">Expression Evaluated To FALSE at line: 23</failure>
     </testcase>
     <testcase classname="test.test_Jenkins" name="test_ignore_message">
            <skipped type="TEST IGNORED">this is stupid, isn't it?</skipped>
     </testcase>
</testsuite>
```

Could you please check my modifications on your environment and give me feedback? Thanks

jeremyhannon:
nope, it fails.

```
22:05:27 Parsing unit test output files into JUnit for Jenkins.
22:05:27 E:/sar/artifacts/jci_unit_testing/master_pr_46_testFarrrbParseChanges/7.0.0.204/Scripts/unity/auto/parse_output.rb:211:in `+': no implicit conversion of nil into String (TypeError)
22:05:27 	from E:/sar/artifacts/jci_unit_testing/master_pr_46_testFarrrbParseChanges/7.0.0.204/Scripts/unity/auto/parse_output.rb:211:in `process'
22:05:27 	from E:/sar/artifacts/jci_unit_testing/master_pr_46_testFarrrbParseChanges/7.0.0.204/Scripts/unity/auto/parse_output.rb:225:in `block in <main>'
22:05:27 	from E:/sar/artifacts/jci_unit_testing/master_pr_46_testFarrrbParseChanges/7.0.0.204/Scripts/unity/auto/parse_output.rb:221:in `each'
22:05:27 	from E:/sar/artifacts/jci_unit_testing/master_pr_46_testFarrrbParseChanges/7.0.0.204/Scripts/unity/auto/parse_output.rb:221:in `<main>'
```

farrrb:
@jeremyhannon thank you for your test. I added an explicit cast (which i forgot in the first place) the code should now be running on your system.

jeremyhannon:
I finally figured out an issue with our setup. Yes now the script successfully runs and Jenkins can parse the `report.xml` again, so things are stable. But for us the `testsuite` `name` field is empty, which looks like a bug.

`<testsuite name="" tests="7" failures="0" skips="0">`

I'm not familiar with Ruby, but in function `process`, it looks like variable `test_suite` isn't initialized before [being used](https://github.com/ThrowTheSwitch/Unity/blob/master/auto/parse_output.rb#L211) (unless its a global).

farrrb:
Can you provide me your output text file, so i can take a look into this?

jeremyhannon:
[UnitTestResults.txt](https://github.com/ThrowTheSwitch/Unity/files/1749931/UnitTestResults.txt)
note: it's output in unity_fixture style.

jeremyhannon:
ok I found the issue. We were passing in an empty string as `argv[0]` which represents the filename, so that must be why it was empty.

```
static const char* argv[] = { "", "-v" };
return UnityMain(2, argv, RunAllTests);
```

farrrb:
Thank you. I'm currently revising the whole parser again, because i found further issues with the script. Our cross compiler for example generates *nix paths on a windows machine and so on... I'll use your output to verify it. It would also be very helpful if you have an output file with errors / ignores for me.

The latest version: https://github.com/farrrb/Unity/blob/fix-parse.rb/auto/parse_output.rb

jeremyhannon:
[UnitTestResults.txt](https://github.com/ThrowTheSwitch/Unity/files/1754714/UnitTestResults.txt)
Here's a version of the same unit tests but with an ignore and a failure.

farrrb:
Hey Jeremy, thank you very much for your input. I took a look at this and refactored the whole parsing script. Here is what i learned:
- There are 3 output formats to parse
  1. "Normal" output
  2. Fixture output
  3. Fixture output with verbose enabled ("-v")
- All three formats are very different
So after a while i ended up refactoring the whole script to make it more readable / extensible... here is my result:
https://github.com/farrrb/Unity/blob/fix-parse.rb/auto/parse_output.rb

I would appreciate it if you could give it a test run. After that i'll file a pull request for that.

Thanks in advance, 
Fab

jeremyhannon:
Ok I tested it in our environment and nothing breaks. I see that the `name` is now hardcoded to `Unity`. Seems ok I guess. Seems good enough to propose as a PR.

farrrb:
Thank you for testing! I took a look at the other scripts and they also hardcoded the name property to "unity" or "Unity" so i decided to do it as well for consistency reasons. Apparently there is no way to get the real "name", because there is none.

PR #321, opened 2018-02-10
MISRA - param names match; macro params in parens

jeremyhannon:
Fix a few small MISRA non-compliance issues:
* parameter name should match between declaration and definition.
* parentheses around all macro parameters


jeremyhannon:
sorry, my first commit failed so I was breaking it into several smaller commits to narrow down the build failure. I'll let you know when it's all staged.

PR #322, opened 2018-02-10
MISRA 19.10: surround each macro param with parens.

jeremyhannon:
MISRA 2004 rule 19.10: inside macros, surround each parameter usage with parentheses.


jeremyhannon:
@mvandervoord ready to merge. My error was trying to surround `__VA_ARGS__` with parentheses.

farrrb:
Just out of curiousity... which MISRA checker do you use? I really like your improvements on the codebase compliance to MISRA C is always a good thing for projects dealing with embedded code.

jeremyhannon:
@farrrb I've turned on selected MISRA checks in the IAR compiler. (My products don't have to be MISRA compliant. I just like the faster compile-time feedback for devs.) We've also found some issues using CppCheck but we've never built that into our process.

PR #325, opened 2018-03-08
Bugfixes and refactoring in parse.rb

farrrb:
There script is now able to parse 3 output formats:
- Normal output (raw unity)
- Fixture output (unity_fixture.h/.c)
- Fixture output with verbose flag set ("-v")

Further improvements:
- Fixed some of the old documentation
- Moved global variables to member variables
- Corrected name of the test suite
- Improved the parsing of unity_fixture ("-v") output. Now paths and lines are parsed, too
- Fixed the stdout output for unity_fixture (both modes)
- Adaptive detection of the path format (slashes *nix and backslashes windows) with cross-plattform capability (e.g. parse windows script from linux and vice versa)

@jeremyhannon was so kind to test the code in his production environment and he gave me his output files to improve my testing. Thank you!

### DIFF
--- a/auto/parse_output.rb
+++ b/auto/parse_output.rb
@@ -1,30 +1,38 @@
 #============================================================
-#  Author:   John Theofanopoulos
-#  A simple parser.  Takes the output files generated during the
+#  Author: John Theofanopoulos
+#  A simple parser. Takes the output files generated during the
 #  build process and extracts information relating to the tests.
 #
 #  Notes:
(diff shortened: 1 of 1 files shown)


jeremyhannon:
yes I have tested these changes in our build environment with VS2013 and IAR RX using unity_fixture -v outputs. Jenkins JUnit plugin still correctly interprets the XML file.

PR #333, opened 2018-07-01
feat: Track individual pointers in leak check

drbartling:
List allocated pointers and the file:line where they were allocated
Remove freed pointers from the list
Report number of leaked pointers when a test fails for leaks
Report file:line of first leaked pointer

Why:
When starting to plug testing into existing code, it can be time
consuming to track down where the memory was leaked from.  This should
significantly expidite the process.

ToDo:
Add tests for changes
Refactor


drbartling:
I tested this using some tests that leaked memory, tried to free memory that was never allocated and tests that freed twice.

mvandervoord:
It appears we're failing because you made use of printf, which we purposefully avoid using in Unity (because some small embedded processors can't take the overhead). You can refactor the printing part to use Unity's print functions.

mvandervoord:
This was never completed or tested and since has been moved around. We can open a new PR to tackle this down the road if desired.

PR #334, opened 2018-07-18
Flush unity output before a longjmp

Xenoamor:
Flush the unity stdout buffer before calling TEST_ABORT().
This is because if TEST_PROTECT() has not previously been called this will cause a segmentation fault and the stdout buffer will fail to print

Although the segmentation fault will still occur, the error that caused it will at least be displayed

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -14,8 +14,8 @@ void UNITY_OUTPUT_CHAR(int);
 #endif
 
 /* Helpful macros for us to use here in Assert functions */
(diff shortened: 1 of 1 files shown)


mvandervoord:
Sorry about the false failure on Travis. I've pushed a change to master that will fix it. If you can merge or rebase master to this PR, we should get a valid test.

Xenoamor:
No worries, these things happen!
I've rebased it for you

PR #356, opened 2018-09-20
UnityPrintNumber(): fix undefined behaviour when printing INT64_MIN

pacheco017:
Simply "negating" a signed value for computing its absolute value does not
always work.

Eg.: given these definitions:
    #define INT64_MIN   (-0x7fffffffffffffff - 1)
    #define INT64_MAX   0x7fffffffffffffff

One can see that -INT64_MIN would overflow. As UnityPrintNumber() negates a
signed variable, this leads to undefined behavior.

The fix is just to special-case MIN values to avoid overflows.

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -181,7 +181,11 @@ void UnityPrintNumber(const UNITY_INT number_to_print)
     {
         /* A negative number, including MIN negative */
         UNITY_OUTPUT_CHAR('-');
-        number = (UNITY_UINT)(-number_to_print);
+        if (number_to_print == UNITY_INT_MIN) {
+            number = UNITY_INT_MIN_ABS;
+        } else {
+            number = (UNITY_UINT)(-number_to_print);
+        }
     }
     UnityPrintNumberUnsigned(number);
 }
--- a/src/unity_internals.h
+++ b/src/unity_internals.h
@@ -99,6 +99,9 @@
     typedef signed char     UNITY_INT8;
     typedef signed short    UNITY_INT16;
     typedef signed int      UNITY_INT32;
+
+    #define UNITY_INT32_MIN INT_MIN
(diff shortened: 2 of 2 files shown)


mvandervoord:
While I like the spirit of this patch, this unfortunately doesn't work for many of the users of Unity. Unity supports many embedded compilers which don't have INT_MAX defined, for example, so we can't add an additional dependency on having that defined.

I'm a little unclear on why you think the definition of INT64_MIN is a problem? -0x7fffffffffffffff - 1 becomes 0x8000000000000000 which IS the smallest value which is able to be represented by a uint64. So it works as intended, does it not? I don't think this falls into undefined behavior in the C specification.

jlindgren90:
@pacheco017 is correct, negating INT_MIN (or INT64_MIN) results in overflow of a signed integer type, which is undefined behavior.

Compiling and running UnityPrintNumber(INT_MIN) with GCC's sanitizers enabled results in a runtime error:
`runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself`

There is an easier fix, though (which the GCC diagnostic hints at).  Replace the line:
`        number = (UNITY_UINT)(-number_to_print);`
With:
`        number = -number;`

This does the negation on an unsigned type, which is well-defined.

jlindgren90:
https://github.com/ThrowTheSwitch/Unity/pull/375

mvandervoord:
Fixed more cleanly in PR #375. Thanks, everyone, for the effort!

pacheco017:
@mvandervoord: I'm really sorry I left you hanging, I've totally missed your first response.

@jlindgren90: Thanks a lot for that fix, it's really simpler and nicer; I'm a little embarrassed I didn't think of doing that ;)

As of today, I've started using your version of the fix and it works like a charm. Thanks a lot for your great work!

PR #368, opened 2018-11-08
Macros in testrunner generator

### LINKED ISSUE #367: Use macros in test runner generator
Hi,
My idea is to change functions in generate_test_runner.rb to macros to give possibilities of (for example) providing output data in another format.
There are two important functions in unity framework: UnityBegin and UnityEnd. They can be redefined using:
https://github.com/ThrowTheSwitch/Unity/blob/master/src/unity_internals.h#L685
and for UnityEnd:
https://github.com/ThrowTheSwitch/Unity/blob/master/src/unity_internals.h#L689

There are functions used in test runner generator:
https://github.com/ThrowTheSwitch/Unity/blob/master/auto/generate_test_runner.rb#L368
https://github.com/ThrowTheSwitch/Unity/blob/master/auto/generate_test_runner.rb#L382

I think it is change easy-to-do. What do you think about it? Is it affecting other parts of the Unity code?
Thanks!

kl-cruz:
I changed UNITY_BEGIN macro to support new parameter. However main change is to provide macro support in testrunner generator.
fixes #367

### DIFF
--- a/auto/generate_test_runner.rb
+++ b/auto/generate_test_runner.rb
@@ -376,7 +376,7 @@ class UnityTestRunnerGenerator
       output.puts('{')
     end
     output.puts('  suite_setup();')
-    output.puts("  UnityBegin(\"#{filename.gsub(/\\/, '\\\\\\')}\");")
+    output.puts("  UNITY_BEGIN(\"#{filename.gsub(/\\/, '\\\\\\')}\");")
     if @options[:use_param_tests]
       tests.each do |test|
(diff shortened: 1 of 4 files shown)


kl-cruz:
Rebased on top of master branch

mvandervoord:
Unfortunately, I can't accept a PR which is going to break the existing `UNITY_BEGIN()` call for existing users. 

Please feel free to submit a new PR if you choose to resolve this. Thanks for the help!

Issue #328, opened 2018-03-26
Runner generator doesn't allow string literals


austinglaser:
A parametrized test of the format

```C
TEST_CASE("hi")
void test_something(char const * str)
{
    TEST_ASSERT_EQUAL_STRING("hi", str);
}
```

does not compile.

The culprit appears to be [this line](https://github.com/ThrowTheSwitch/Unity/blob/master/auto/generate_test_runner.rb#L95) of `find_tests()` in `generate_test_runner.rb`, where literal strings are scrubbed out of the source.

I understand (I think) the point of removing string literals, since they have the potential to contain characters that will confuse later regex parsers.  However, there's got to be a better way to go about this. At a minimum, the current scrubbing regex doesn't handle strings containing quotes -- for instance:

```C
TEST_CASE("{\"field\": 1234}")
void test_handles_json(char const * json)
{
    /* ... */
}
```

will thoroughly break the parser.

art-of-dom:
This should be closed as #380 fixed this

Issue #282, opened 2017-05-03
Support strings in TEST_CASE()


rmja:
It would be neat if the TEST_CASE() macro supported string arguments, such as:

```
TEST_CASE( "Test string" )
void test_string( const char* aString )
{
}
```
[This line](https://github.com/ThrowTheSwitch/Unity/blob/master/auto/generate_test_runner.rb#L95) removes the string, so the generated `main()` in the test runner becomes:

```
int main(void)
{
  UnityBegin("test_parameterized.c");
  RUN_TEST(test_string, 37,   );

  return (UnityEnd());
}
```

instead of the desired

```
int main(void)
{
  UnityBegin("test_parameterized.c");
  RUN_TEST(test_string, 37,  "Test string" );

  return (UnityEnd());
}
```

By removing the line, things seems to work, but I guess there is a reason why the line is there in the first place:)

rmja:
Note that my simple remove-the-line fix does not work nice with the case where a `/` is in the `TEST_CASE` string.

SyntaxColoring:
@mvandervoord I think this request can be closed.  It was implemented by PR #380.

mvandervoord:
Thanks!

PR #380, opened 2018-12-10
Add support for strings in TEST_CASE()

### LINKED ISSUE #361: Auto test generatetor scrubs string in TEST_CASE
I've been trying to utilize the test cases more in my unit tests, but I have run into an issue when using strings in test cases. The auto test runner seems to strip out any string.

For example, a test that looks like this:
```
TEST_CASE(0, "SUCCESS")
void test_Foo_ReturnValueMapsToString(int status, const char * str)
{
    TEST_ASSERT_STRING(foo(status), str);
}
```

Would result in the follow (which obviously wont even compile),
`
RUN_TEST(test_Foo_ReturnValueMapsToString, 0, );
`

I've tracked down the offending line which is [line 96 of the generate test runners](https://github.com/ThrowTheSwitch/Unity/blob/master/auto/generate_test_runner.rb#L96). When I comment out this line, it generates code with the string. 

I'm happy to do a PR to help fix this, but two questions:

1. What is the original intent of this line? It doesn't seem to affect any of the tests outside TEST_CASE across many test files.
2. If we still need to keep this line, is there any case outside the TEST_CASE we should keep the string in tact?

elliot-gawthrop:
Modified generate_test_runner to support strings in parameterized tests

Fixes #361 #328 #282

### DIFF
--- a/auto/generate_test_runner.rb
+++ b/auto/generate_test_runner.rb
@@ -90,12 +90,25 @@ class UnityTestRunnerGenerator
   def find_tests(source)
     tests_and_line_numbers = []
 
+    # contains characters which will be substituted from within strings, doing
+    # this prevents these characters from interferring with scrubbers
+    # @ is not a valid C character, so there should be no clashes with files genuinely containing these markers
+    substring_subs = { '{' => '@co@', '}' => '@cc@', ';' => '@ss@', '/' => '@fs@' }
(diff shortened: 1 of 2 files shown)


mvandervoord:
Thanks @elliot-gawthrop ! You rock!

PR #386, opened 2019-01-26
Fix error in example_1

teaguecl:
This test case had an error:
test_FindFunction_WhichIsBroken_ShouldReturnZeroIfItemIsNotInList_WhichWorksEvenInOurBrokenCode
It was supposed to be a list of values that are NOT in the list,
and none of them should be found.  It incorrectly included '1'
which is a value in the list.

The compile option -Wno-misleading-indentation was also added to
remove a compiler warning produced by gcc 7.3.0


teaguecl:
I was confused looking at the examples because it did not behave as described.

Note my editor is set up to trim trailing whitespace, which lead to a little more delta that I expected.  The change is for the better I think, so I left it in.

Issue #392, opened 2019-03-01
Array comparison functions are incorrect what CHAR_BIT is 16


detly:
I am building and running Unity on a system where `CHAR_BIT == 16` (not uncommon on DSPs). Amongst other things, this means that `sizeof(int16_t) == sizeof(int) == sizeof(char) == 1` Some of my tests use `TEST_ASSERT_EQUAL_INT16_ARRAY()`. I found that this spuriously fails certain tests because it runs over the end of the input arrays.

The problem seems to come down to:

  1. When the `INT16` variant is used, the style is set to `2 + (something else)`. The `(something else)` is irrelevant here.
  2. At the start of [UnityAssertEqualIntArray()](https://github.com/ThrowTheSwitch/Unity/blob/master/src/unity.c#L812), `length` is set to `2` [by masking](https://github.com/ThrowTheSwitch/Unity/blob/b723c9f25012155b34d698094155164f465a0412/src/unity.c#L821) the `(something else)` out of `style`.
  3. At the end of the loop, the pointers to each array are incremented by `length` ie. 2. This is incorrect on this system, an array of `int16_t` should only be incremented by 1 (since it's already the smallest addressable unit). I've verified in a debugger that adjacent members of `int16_t` arrays have addresses that only differ by one.

I don't think there are any modifications to `unity_config.h` parameters that will help with this. I'm also at a loss as to how you'd test this on "normal" hardware. Perhaps with a simulator for such a system?

detly:
Here's a minimal example. Save this as `main.c` with Unity in a `unity` directory alongside.

```c
#include <inttypes.h>
#include "unity/src/unity.h"

static void test(void)
{
    const int16_t actual[4] = {1, 1, 0, 0};
    const int16_t expected[4] = {1, 1, 1, 1};
    TEST_ASSERT_EQUAL_INT16_ARRAY(expected, actual, 2);
}

int main(void) {
    UNITY_BEGIN();
    RUN_TEST(test);
    return UNITY_END();
}
```

On my desktop system:

```text
$ gcc -o test main.c unity/src/unity.c
$ ./test 
main.c:13:test:PASS

-----------------------
1 Tests 0 Failures 0 Ignored 
OK
```

On the hardware (via an IDE):

```text
../main.c:8:test:FAIL: Element 1 Expected 1 Was 0

-----------------------
1 Tests 1 Failures 0 Ignored 
FAIL
```

detly:
[...]

PR #394, opened 2019-03-04
CHAR_BIT ≠ 8 platform support

detly:
Before reviewing this, please see issue #392 and PR #393 . In particular, this contains all the changes in that PR, so I won't explain them again here unless asked.

The changes here decouple the character size from the display size for all of the integral types. It mostly comes down to the `UNITY_OCTET_PER_CHAR` constant which is just `ceil(CHAR_BIT/8)`. Anything involving displaying an integral type in hexadecimal uses this constant to convert the platform sizes to a number of nibbles for display.

So for example, on a platform where `CHAR_BIT` is 8, everything works as it did before. If `CHAR_BIT` is 16 (or anything from 9-16 in fact) then array comparison functions use the "size in `char`s" for incrementing pointer addresses, while display functions use "size in octets" for counting nibbles.

Note that the named types *above the platform byte size* are still accurate eg. `UNITY_INT16` is a 16-bit integer. On a `CHAR_BIT == 16` platform it has a size of 1, comprises 2 octets, and requires 4 nibbles to print. **However** `UNITY_INT8` exists but is 16 bits. There's not really any way around this. The alternative seemed to be disabling the type altogether, which would have much more of an impact on the interface and lead to all sorts of ugliness. This way the type still exists, and behaves in most of the same ways as it would on other platforms.

The only quirk here is that if `CHAR_BIT > 8`, some functions more or less "ignore" certain cases and data types eg. in `UnityAssertEqualIntArray()` or `UnityNumToPtr()` the `case 1:`s will never happen. I don't see this as a problem, since the functions themselves still do what they're meant to.


mvandervoord:
Don't worry about the Ci failure. It's related to versioning of the test system and will get fixed.

I'm not seeing how the variadic macros changes are related to this PR.

detly:
> I'm not seeing how the variadic macros changes are related to this PR.

Only tangentially — I can't run Unity's own unit tests on the TI hardware without enabling the variadic version of `RUN_TEST(...)`.

On that note, I forgot to mention:

- All tests (including `printcharSpy`-related ones) pass on my amd64 PC.
- All tests except two (including `printcharSpy`-related ones) pass on my TI development board where `CHAR_BIT` is 16. The two failures are related to printing floats: `-0.0f` and `-nan` which I strongly suspect are simply TI's compiler converting the compile-time constant expressions to `0.0f` and `nan` before the test function even sees it.

mvandervoord:
Replaced with #403 and #404. thanks!

Issue #398, opened 2019-03-13
Multi line test-function signatures


frot:
Ceedling/Unity currently can't handle test-function signatures spanning several lines when using parameterized tests.
Most IDEs and editors will split the function parameters on several lines and this will not work with unity.

> test_SomeTestcaseFunction(int param_a,
>        int param_b,
>        int param_c)

PR #399, opened 2019-03-13
Allow multi line test-function definitions.

frot:
Ceedling/Unity currently can't handle test-function signatures spanning several lines when using parameterized tests. See #398


dpostorivo:
This build should be rebased on master for the CI fixes.

PR #403, opened 2019-04-06
Used sizeof operator for pointer increments in UnityAssertEqualIntArray().

farrrb:
This pr should fix issue #392. 

The problem with the TI DSPs (e.g. the ugly C2000 series) is that the increment for a 16-bit type is 1 instead of 2. By using the sizeof operator for (all) pointer increments this problem can be circumvented.
There is also problem when using sizeof in comparisons and calculations (e.g. for masks). This is fixed by taking the bit width of UNITY_INT and divide it by 8 to get the real number of bits.

**Additional changes:**
- enclosed one "if-return 0" statement in curly braces
- updated the copyright date to 2019
- formatted enum values (UNITY_DISPLAY_STYLE_INT and UNITY_DISPLAY_STYLE_UINT)

This pr goes along with #404.

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -1,6 +1,6 @@
 /* =========================================================================
     Unity Project - A Test Framework for C
-    Copyright (c) 2007-14 Mike Karlesky, Mark VanderVoord, Greg Williams
+    Copyright (c) 2007-19 Mike Karlesky, Mark VanderVoord, Greg Williams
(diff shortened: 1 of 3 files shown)


detly:
Does this need to be extended to the functions added in #412 ?

farrrb:
> Does this need to be extended to the functions added in #412 ?

Yes absolutely!

mvandervoord:
Replaced by #425

PR #404, opened 2019-04-07
Fixed maximum number of nibbles for processor where sizeof() operator…

farrrb:
… doesn't return the size of a type in 8-bit bytes (e.g. the TI C2000 series). This patch makes the code more explicit and readable. 

See also: #394 #392 #403

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -352,9 +352,9 @@ void UnityPrintNumberHex(const UNITY_UINT number, const char nibbles_to_print)
     int nibble;
     char nibbles = nibbles_to_print;
 
-    if ((unsigned)nibbles > (2 * sizeof(number)))
+    if ((unsigned)nibbles > UNITY_MAX_NIBBLES)
     {
-        nibbles = 2 * sizeof(number);
+        nibbles = UNITY_MAX_NIBBLES;
     }
 
     while (nibbles > 0)
--- a/src/unity_internals.h
+++ b/src/unity_internals.h
@@ -122,18 +122,20 @@
  * 64-bit Support
  *-------------------------------------------------------*/
 
+/* Auto-detect 64 Bit Support */
 #ifndef UNITY_SUPPORT_64
   #if UNITY_LONG_WIDTH == 64 || UNITY_POINTER_WIDTH == 64
     #define UNITY_SUPPORT_64
   #endif
 #endif
 
+/* 64-Bit Support Dependent Configuration */
 #ifndef UNITY_SUPPORT_64
     /* No 64-bit Support */
     typedef UNITY_UINT32 UNITY_UINT;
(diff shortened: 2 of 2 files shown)


detly:
Is this meant to go along with my PRs to address a similar issue, or did we arrive at this independently?

farrrb:
Hey @detly I'm sorry I forget to write you a note (my toddler needed some attention and then I forgot...). This and pr #403 are an alternative way of fixing the problems you found in **UnityPrintNumberHex** & **UnityAssertEqualIntArray**.

After seeing your PRs I revisited all the places sizeof() was used in the code and tried to fix the problems they bring with it.

I saw in your application that you used **CHAR_BIT** by default which is provided by **limits.h** of the standard library. This is an optional (configurable) include and Unity should work without it, so I decided to tryout a fix without this header in mind.
 
Another thing I didn't agree with, is putting sizeof()-lengths in Unitys interface. 
In my past experience dealing with the TMS320F281x it was always better to treat the interfaces as they were byte oriented and to do the byte-juggling on the inside of the functions. This avoided a lot of headaches when revisiting our code.

Please don't take this as an offence, it is just another approach. I'd really like to get some TMS320F281x support for Unity in the future and I think we want the same thing ;)

detly:
Yeah that's fine! I had forgotten about the dependency on `limits.h`. I'll leave it to the maintainer to figure out which approach they want to use.

mvandervoord:
@detly & @farrrb -- Thank you both for the hard work! If I'm understanding correctly, #403 & #404 covered the main features you were interested in, yes? If I've missed anything from #394 I apologize. Please feel free to create a new PR with what you feel should be included yet. 

Again, thank you both so much. Our vision for these tools has always been to be able to support all the quirky micros out there and I really appreciate your help in getting it there!

detly:
@mvandervoord I'll test it some time in the next few weeks. It does look like it drops the variadic macro fix included in PR #393 , and I'm not sure about the pointer width restriction stuff too. You should also check my comment about Unity just not running its own tests on strict C90 compilers that don't support variadic macros.

farrrb:
Yes please revisit your proposal @detly. I included only the things I didn't agree with (implementation not the idea itself). 

I suggest you open a different PR for every feature I did not include, because I really liked your ideas.
- UNITY_SUPPORT_8 (maybe with a solution to enforce 8 bit support for platforms without limits.h or with inverted logic like UNITY_DISABLE_SUPPORT_8)
- relaxed pointer widths
- the variadic macro fix

detly:
> * UNITY_SUPPORT_8 (maybe with a solution to enforce 8 bit support for platforms without limits.h or with inverted logic like UNITY_DISABLE_SUPPORT_8)

I like this idea, I'll make that change.

PR #406, opened 2019-04-09
Unity framework Typecast warning issues solved

jaiqc:
UnityAssertEqualNumber unsigned input argument is signed overlap issues solved

eg:
#include <stdio.h>
#include <stdint.h>

void UnityAssertEqualNumber (int32_t a)
{
    printf("%d", a);    
}

int main()
{
    fUnityAssertEqualNumber((int32_t)(uint32_t) 0xFFFFFFFF);
    return 0;
}

Output: -1


mvandervoord:
Between the conflicts and the many many failing self-tests, I can't merge this in this state. Feel free to open a new PR if you come to a cleaner solution. Thanks for helping out!

Issue #405, opened 2019-04-07
What about a coding style / code formatter for this project?


farrrb:
Everytime I do a change for this project I try to be a good boy-scout and leave the place a little cleaner. 😄 But I recently noticed that the indentation of the preprocessor macros for example is totally off. 
So here's my proposal: how about adding a code formatter configuration file (e.g. for clang format) in order to make sure the code is according to style before we do commits?

mvandervoord:
:) Using Rubocop to check style of both ruby and C.

austinbes:
Out of curiosity, what configuration do you use for Rubocop and C?

mvandervoord:
I apologize. I've misspoken. 

The Ruby code is style-checked by Rubocop (you can find the style configuration in `test/.rubocop.yml`)

The C code is style-checked by AStyle. It's settings are in the `test/rakefile_helper.rb` (and currently are as follows):

- allman
- indent=spaces=4
- indent-switches
- indent-preproc-define
- indent-preproc-block
- pad-oper 
- pad-comma
- unpad-paren
- pad-header
- align-pointer=type
- align-reference=name
- add-brackets
- mode-c
- suffix=none

austinbes:
Awesome, thanks!

PR #415, opened 2019-05-06
Refactoring spaces and braces / Enforcement of style

farrrb:
This is a really large commit which enforces style and format on all C-source and C-header files based on the de-facto standards present in the code base.
I carefully crafted these individual commits with clang-format and did some modifications by hand in order to correct and/or revert single changes.

Additionally I added some braces where needed and corrected some include guards (name and formatting).

This pr is crafted to support the request of #405. Please review _all_ the changes carefully in case I missed something.


extras/fixture/src/unity_fixture.c
```diff
@@ -151,7 +159,7 @@ static int malloc_fail_countdown = MALLOC_DONT_FAIL;
 
 void UnityMalloc_StartTest(void)
 {
-    malloc_count = 0;
+    malloc_count          = 0;
```
Letme: This is strange. I understand where you are filling a struct you would want to have `=` aligned, but here it is a separate variable...

extras/fixture/src/unity_fixture.c
```diff
@@ -178,7 +186,7 @@ void UnityMalloc_MakeMallocFailAfterCount(int countdown)
 
 #ifdef UNITY_EXCLUDE_STDLIB_MALLOC
 static unsigned char unity_heap[UNITY_INTERNAL_HEAP_SIZE_BYTES];
-static size_t heap_index;
+static size_t        heap_index;
```
Letme: again the alignment here I doubt it is what we expect/have elsewhwere?

extras/fixture/src/unity_fixture.c
farrrb: Actually there were both types in the code base if you look closely:
- https://github.com/ThrowTheSwitch/Unity/blob/c9b1d217fd0cb26929d3554da1377e90aa583198/src/unity.c#L1015-L1018
- https://github.com/ThrowTheSwitch/Unity/blob/c9b1d217fd0cb26929d3554da1377e90aa583198/src/unity.c#L1091-L1093

I decided to take the one that looked cleaner to me and since we aligned a lot of stuff in the code it also seemed more consistent to use this alignment here as well.

extras/fixture/src/unity_fixture.c
farrrb: See above.

mvandervoord:
I'm afraid I have let this branch diverge too much before merging it. I am very sorry about that. I suspect that it would be easiest to do as a new PR at this point.

PR #416, opened 2019-05-08
UnityAssertEqualNumber unsigned input argument is signed overlap issues solved

jaiqc:
eg:
#include <stdio.h>
#include <stdint.h>

void UnityAssertEqualNumber (int32_t a)
{
printf("%d", a);
}

int main()
{
fUnityAssertEqualNumber((int32_t)(uint32_t) 0xFFFFFFFF);
return 0;
}

Output: -1


mvandervoord:
It would appear that something went wrong when you tried to do your merges. Sorry, git can definitely be painful when this sort of thing happens! It would probably be easier to start with ThrowTheSwitch.org's master version and then cherry-pick your changes back into it and try a new PR.

Issue #421, opened 2019-05-09
Unity test testHexPrintsUpToMaxNumberOfNibbles() fails on TI C55x


detly:
This is the test:

```c
void testHexPrintsUpToMaxNumberOfNibbles(void)
{
#ifndef USING_OUTPUT_SPY
    TEST_IGNORE();
#else
    startPutcharSpy();
    UnityPrintNumberHex(0xBEE, 21);
    endPutcharSpy();
    TEST_ASSERT_EQUAL_INT(sizeof(UNITY_INT)*2, strlen(getBufferPutcharSpy()));
#endif
}
```

On TI, we have:

* `UNITY_MAX_NIBBLES == 8`
* `UNITY_INT_WIDTH == 16`
* `sizeof(UNITY_INT) == 1`
* actual `strlen`: 8
* expected: 4

So, fail.

For comparison, on an amd64 PC we have:

* `UNITY_MAX_NIBBLES == 16`
* `UNITY_INT_WIDTH == 64`
* `sizeof(UNITY_INT) == 8`
* actual `strlen`: 16
* expected: 16

Pass.

I cannot figure out how to make this test pass consistently on all platforms without introducing a dependency on `CHAR_BIT` ie. `limits.h`.

detly:
Possible alternatives:

- simply have the expected value be `UNITY_MAX_NIBBLES`, although I'm not sure how useful a test that is (you'd almost just be checking that the preprocessor does division).

- add a case for `UNITY_MAX_NIBBLES` to be 4 on platforms like this, or just `UNITY_INT_WIDTH/4` in general.

farrrb:
UNITY_INT should be at least 32-bit wide on all platforms. 

https://github.com/ThrowTheSwitch/Unity/blob/c9b1d217fd0cb26929d3554da1377e90aa583198/src/unity_internals.h#L131-L132

https://github.com/ThrowTheSwitch/Unity/blob/c9b1d217fd0cb26929d3554da1377e90aa583198/src/unity_internals.h#L145-L146

So my question is:
Isn't that the case for the TI C55x? So **sizeof**(UNITY_INT) should be **2** in your example (correct me please if I'm wrong). And in that case I would expect **8** for **UNITY_MAX_NIBBLES**. 

Please correct me if I'm not getting everything.

Imho the test is a good idea, but I would prefer changing the line to:
```C
void testHexPrintsUpToMaxNumberOfNibbles(void)
{
[...]

Issue #420, opened 2019-05-09
Pointer comparison fails on TI C55x


detly:
On the TI C55x using their CGT compiler, pointer widths are 23 bits, whereas `int`s and `char`s are 16 bits. So `sizeof(int *)` is 2, and the pointer gets truncated to 16 bits in `UnityNumToPtr()`'s `case 2:`.

My previous PR for this had a fix, but it's no longer compatible with what was in PR #403 , and I can't see another way to resolve it.

(Note that `UnityNumToPtr()` is *always* going to be using the wrong size for such a platform eg. `UNITY_TEST_ASSERT_EACH_EQUAL_INT16()` will call `UnityNumToPtr(expected, 2)` when the real size is 1. But the pointer case is the only case where the incorrect size is too small rather than too big.)

farrrb:
Can you provide a fully featured example code? 

UnityNumToPtr() shouldn't truncate the pointer at all, only the expected value should be truncated (upper bits). You put a 32 (or 64-bit) UNITY_INT into the function and get a pointer to the truncated version in the quick compare structure as a result. Am I missing something?

detly:
The example code is just Unity's own unit test:

```text
tests/testunity.c:4623:testEqualPtrEachEqual:FAIL: Element 0 Expected 0x71FC0000 Was 0x000171FC: [[[[ Test Should Have Passed But Did Not ]]]]
```

The test is:

```c
void testEqualPtrEachEqual(void)
{
    char A = 1;
    char B = 2;
    char C = 3;
    char* p0[] = {&A, &A, &A};
    char* p1[] = {&A, &B, &C, &A};
    char* p2[] = {&B, &B};
    char* p3[] = {&C};

    TEST_ASSERT_EACH_EQUAL_PTR(&A, p0, 1);
    TEST_ASSERT_EACH_EQUAL_PTR(&A, p0, 3);
    TEST_ASSERT_EACH_EQUAL_PTR(&A, p1, 1);
    TEST_ASSERT_EACH_EQUAL_PTR(&B, p2, 2);
    TEST_ASSERT_EACH_EQUAL_PTR(&C, p3, 1);
}
```

[...]

Issue #418, opened 2019-05-09
Unity's own tests won't compile if variadic macros aren't supported


detly:
If you try to compile and run Unity's own tests on a compiler that doesn't support variadic macros, they won't compile eg. on TI's CGT clxx:

```
build/testunityRunner.c: In function ‘main’:
build/testunityRunner.c:464:53: error: macro "RUN_TEST" passed 2 arguments, but takes just 1
     RUN_TEST(testUnitySizeInitializationReminder, 86);
                                                     ^
```

(Note that this compiler DOES support variadic macros but Unity doesn't know that, so attempts to use non-variadic forms. It's just for illustration. You should be able to reproduce this with any compiler in strict ANSI mode.)

detly:
This is still an issue. It bites me every time I update to see if Unity has upstream support for TI.

mvandervoord:
Which target are you specifying when you try to run the self tests? 

There isn't one in the list for TI c55x, but if we add one, clearly we wouldn't want to run the tests for parameterized tests because the compiler wouldn't support it. 

Would this solve your problem or am I missing the point?

detly:
> Which target are you specifying when you try to run the self tests? 

I'm not sure what you mean by this.

The TI compiler can be set to support variadic macros or not to.

If I set it not to (strict ANSI mode), the tests straight up don't compile because the support for non-variadic macro operation seems to be broken (I think... it's complicated to set this up for compilation by CCS).

If I set it to support variadic macros (allow GCC extensions + relaxed parsing), they still don't compile because the `__STDC_VERSION__ >= 199901L` check doesn't perfectly map to "supports variadic macros" - the TI compiler doesn't set it.

So one bug is: if Unity intends to support running its own tests on targets without support for variadic macros, it doesn't* which is a bug. (*again, I'm not 100% sure about this, so feel free to verify)

Another bug is: when compiling Unity's own tests on a compiler that supports variadic macros but doesn't set `__STDC_VERSION__ >= 199901L`, variadic macros are not used, which should probably also be a bug. It's fixed by [this](https://github.com/detly/Unity/commit/138dc8175d2d75a1e5c5f408bcdf1e8b277742ad) and [this](https://github.com/detly/Unity/commit/132261d6146a39be9a3fcac48c84e6fb3ea55d27) if you're interested.

mvandervoord:
Thanks for the explanation and the support. Fixed.

Issue #417, opened 2019-05-09
Impossible to not have setup stubs defined


detly:
When building Unity's own tests for the TI using their C90-ish compiler, I get the error:

```text
<Linking>
error #10056: symbol "_tearDown" redefined: first defined in "./src/unity.obj";
   redefined in "./test/tests/testunity.obj"
error #10056: symbol "_setUp" redefined: first defined in "./src/unity.obj";
   redefined in "./test/tests/testunity.obj"
error #10010: errors encountered during linking; "Unity.out" not built
```

This is because both:

* `unity.c` defines `UNITY_INCLUDE_SETUP_STUBS` immediately before including `unity.h`, so we get the empty definitions
* `testunity.c` defines its own set

There seems to be no way to avoid this collision.

`unity.c` has a hardcoded `#define UNITY_INCLUDE_SETUP_STUBS` immediately before including `unity.h`, so it's permanently on. TI's compiler has no concept of a `weak` function attribute, so there's no sidestepping it that way, the only logical thing that can be done is to use `--define=UNITY_WEAK_ATTRIBUTE=""`. But that still gives a collision.

I don't know enough about the need for this guard to make the fix myself.

PR #422, opened 2019-05-09
TI C55x compatibility

detly:
This includes some of the minor fixes from PR #393 that didn't make it in (tweaked a bit as per discussion):

> Here are some smaller changes that help getting Unity to work more fully on some of TI's DSP platforms. Specifically:

- Relaxed the pointer width restrictions. A few DSPs have weird pointer widths like 23/24 bits (or eg. 15, or 31, due to the top bit being reserved for special "kinds" of addressing). There's no reason you can't cast a 23-bit pointer into a 32-bit int and print that, so I relaxed the restrictions on this.

- Check `CHAR_BIT`. If it's not exactly 8 (remember, the standard says it must be *at least* 8), there's now an extra macro `UNITY_NO_SUPPORT_8` defined that allows for ignoring tests that rely on strictly-8-bit behavious

- TI's clxx compilers are nominally C89 compilers, but support a bunch of C99 features such as variadic macros. This is actually a common thing in embedded toolchains. So I added a macro and supporting docs to tell Unity to use the variadic macro version of `RUN_TEST()` even on a claims-to-be-C89 toolchain.

Issues I couldn't fix due to incompatibility with PR #403 :

- #420 
- #421 

Other related issues that aren't really that critical:

- #418 
- #417


farrrb:
I'm not sure about the relaxed pointer widths anymore. Maybe unity should auto-detect the width and adjust the value to the next power of 2 in order to do arithmetic on the widths. It is nearly impossible to derive a type in which the pointer fits based on an arbitrary pointer width given (23 for example). 

I think it would be wise to differentiate between the system's pointer width and unity's pointer width (like UNITY_INT which is always 32 or 64 bit but never 16 bits wide).

See #420 for more information.

detly:
Entirely possible, but I guess my problem with that is — what's the point
in calculating an arbitrary power of two if various other parts of Unity
assume 32 or 64 only?

detly:
(Also how would it auto-detect the pointer width in bits? We can't even rely on auto-detecting a `char` width in bits without introducing a dependency on `limits.h`.)

farrrb:
The mechanism for unity is a little bit complicated but it works very well on many platforms.
There are three stages:
1. check limits.h for types and determine the integer (UINT_MAX) & long (ULONG_MAX) type
2. check stdint.h (UINTPTR_MAX) for types and determine the pointer type (with using the next power of two)
If this doesn't work (in case there is no limits.h and/or nor stdint.h):
3. fallback to default (UNITY_INT = 32-bit, UNITY_LONG_WIDTH = 32-bit and UNITY_POINTER_WIDTH = 32-bit). 

This mechanism can be overriden by manual definition of these macros. But you have to keep in mind that you should only set power-of-two values, because the slightly smaller types will be handled like them (which is totally fine in my opinion). 
One should always try to autodetect your processor first. Only if this doesn't work one should set the macros manually. 

The TI C55x should be autodetected with a 16-bit int type, a 32-bit long type and a 32-bit pointer type as I can infer from the limits.h and stdint.h of the C55x standard library. Imho the documentation should be updated to make this more clear.

mvandervoord:
Replaced by #425

PR #429, opened 2019-07-03
Meson support for Unity unit testing framework.

squidfarts:
## Hello 😀.
I would like to provide support for Meson build system for the Unity unit testing framework.  If you're interested check out their site and documentation.

* [Meson website](https://mesonbuild.com/index.html)
* [Manual](https://mesonbuild.com/Manual.html)
* [Reference](https://mesonbuild.com/Reference-manual.html)

Meson also has a dependency tool that downloads dependencies from third party venders like you.  I happen to be preparing your framework so I can use a "[wrap-git]" file because I love your testing framework.  

* [WrapDB](https://mesonbuild.com/Wrap-dependency-system-manual.html)


mvandervoord:
@squidfarts -- Would it be more valuable to people if the meson builds were in the examples directory so people could see how to USE meson with Unity?

squidfarts:
I well get on that when I get back from work.  Quick question, should I put a Meson.build in all three examples?

mvandervoord:
No rush.

I think I would add a fourth example that looks like the 1st, but is a meson build instead of a make build. :)

squidfarts:
I added a fourth example into the repository pull requests also an 'meson_options.txt'.

PR #432, opened 2019-07-14
Quick fix for Meson build script. Make static library only.

squidfarts:
This marge request now includes a small fix that would make a difference to the end user.  An issue with linking against a shared version of Unity vs a static version of the library.

I wanted to add one more example for C++, simply to show that Unity can be used with C++ projects.


squidfarts:
This marge request now includes a small fix that would make a difference to the end user.

PR #440, opened 2019-09-10
Fix undefined behavior in UnityPrintNumber

jlindgren90:
And add sanitizers to the GCC build flags so that no-one breaks this again.

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -327,7 +327,11 @@ void UnityPrintNumber(const UNITY_INT number_to_print)
     {
         /* A negative number, including MIN negative */
         UNITY_OUTPUT_CHAR('-');
-        number = (UNITY_UINT)-number_to_print;
+        /* Take the two's complement of the unsigned value.  We could use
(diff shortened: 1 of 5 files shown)


jlindgren90:
Is there any particular reason your CI builds are still on GCC 4.8?

mvandervoord:
Awesome. Thank you. 

I must apologize, I just committed the latter portion, but I like yours with the comment. Can you merge master and update the PR?

mvandervoord:
Yes. To catch problems like the one you just committed. :)

I've been meaning to add additional versions so that I can run tests with cutting edge AND older versions... but haven't been able to thus far.

jlindgren90:
You don't need anything cutting-edge to run sanitizers.  GCC 4.9.0 (5+ years old) or even older versions of clang would do it.

I'll close this.  Feel free to copy the comment if you like.

mvandervoord:
4.9 sounds like a winner, then! Thanks for the advice!

Issue #438, opened 2019-09-10
[regression] Undefined references to setUp and tearDown


jlindgren90:
Please revert commit 45020b0d3b03; it causes the following build errors on MinGW when setUp() and tearDown() are not overridden (i.e. when relying on the weak symbols):

```
ld.exe: unity.o: in function `UnityDefaultTestRun': unity.c:1748: undefined reference to `setUp'
ld.exe: unity.c:1753: undefined reference to `tearDown'
collect2.exe: error: ld returned 1 exit status
```

Note that under MinGW, weak symbols need to be defined in the same module that references them.  I made a note of this in https://github.com/ThrowTheSwitch/Unity/blob/2939c420ed9f2697bcb82c1b193bb48c4a8ad7a9/src/unity.h#L33 but it appears to have been overlooked.

The correct fix for #417 would be to make sure that neither UNITY_WEAK_ATTRIBUTE nor UNITY_WEAK_PRAGMA are defined when using the TI C55x compiler.  That's how you're supposed to tell Unity that the compiler doesn't support weak symbols, and in that case, UNITY_INCLUDE_SETUP_STUBS will do nothing.

mvandervoord:
I rejected the PR, but not the need to fix this. I don't love a solution which requires more core files to be includes.

I get that it's almost hypocritical: I spend a lot of time teaching people to break their files into smaller parts and such, but a common complaint that we get with Unity is that it's not all in a single header file. I'd like to not make it MORE files.

I'm not sure I follow the logic in the 2nd statement, that it "does not work if unity.h is included
(directly or indirectly) via a command-line directive." That seems to be a place where it easily DOES work. If UNITY_INCLUDE_SETUP_STUBS is defined at the command line, then it's ALWAYS defined when needed, right? How is that a problem?

We've attempted "optional" `setUp` and `tearDown` a number of ways now, primarily with `weak` functions. Thus far it's been fairly problematic because of the inconsistencies. GCC's support of weak changes from version to version. Other compilers don't support it at all, or support it in a completely different way. What started as a convenience idea for those using gcc has ballooned into a complicated mess of options.

I'm considering dropping all the weak support and returning to a point where setUp and tearDown are always required again. For those using Ceedling and/or the test runner generator, we could have it automatically look to see if the functions exist in each test file, and generate the runner to have replacements if they don't. I think this would solve it in a compiler agnostic way. 

Thoughts?

Letme:
[...]

PR #443, opened 2019-09-13
Introduce unity_setup_stubs.h and eliminate UNITY_INCLUDE_SETUP_STUBS.

jlindgren90:
Requiring a macro like UNITY_INCLUDE_SETUP_STUBS to be defined before
including unity.h is fragile and does not work if unity.h is included
(directly or indirectly) via a command-line directive.

This also resolves issue #438.


mvandervoord:
I understand the motivation with this one, but I don't love a solution which requires more core files to be includes.

I get that it's almost hypocritical: I spend a lot of time teaching people to break their files into smaller parts and such, but a common complaint that we get with Unity is that it's not all in a single header file. I'd like to not make it MORE files.

I'm not sure I follow the logic in the 2nd statement, that it "does not work if unity.h is included
(directly or indirectly) via a command-line directive." That seems to be a place where it easily DOES work. If `UNITY_INCLUDE_SETUP_STUBS` is defined at the command line, then it's ALWAYS defined when needed, right? How is that a problem?

I still want to solve the overall problem. I'm moving the conversation back to issue #443 that discusses this.

mvandervoord:
I rejected the PR, but not the need to fix this. I don't love a solution which requires more core files to be includes.

I get that it's almost hypocritical: I spend a lot of time teaching people to break their files into smaller parts and such, but a common complaint that we get with Unity is that it's not all in a single header file. I'd like to not make it MORE files.

I'm not sure I follow the logic in the 2nd statement, that it "does not work if unity.h is included
(directly or indirectly) via a command-line directive." That seems to be a place where it easily DOES work. If `UNITY_INCLUDE_SETUP_STUBS` is defined at the command line, then it's ALWAYS defined when needed, right? How is that a problem?

I still want to solve the overall problem. I'm moving the conversation back to issue #443 that discusses this.

PR #453, opened 2019-10-22
Refactor/optional setup through scripting

mvandervoord:
@jlindgren90 & @Letme (and anyone else willing to look), could you check this out?

This removes the "weak" support from unity, replacing it with a completely test-runner-generator approach. If you specify setup / teardown and/or suite functions, they are automatically added to the runner, otherwise they are NOT part of the runner and therefore not necessary.


jlindgren90:
@mvandervoord I like it, but there is one problem, which is that UnityDefaultTestRun() still calls setUp() and tearDown() unconditionally.  So when they are not present in the test file, I get linker errors:

```
ld: unity.o: in function `UnityDefaultTestRun':
unity.c:1748: undefined reference to `setUp'
ld: unity.c:1753: undefined reference to `tearDown'
collect2: error: ld returned 1 exit status
```

Maybe the generator script could create an empty setUp() and tearDown() function when they are not present in the test file?

mvandervoord:
@jlindgren90 -- I had originally planned to do it that way, but convinced myself that it wasn't necessary because if they're using generate_test_runner, it's generating a RUN_TEST and therefore UnityDefaultTestRun isn't used... but you're right... it's still needs to be able to LINK. My bad. I'll fix that.

jlindgren90:
@mvandervoord How about this: https://github.com/jlindgren90/Unity/commit/4d231a68efa729fb27c284ae491034b78a4c007b

jlindgren90:
Or slightly simpler: https://github.com/jlindgren90/Unity/commit/68cc45a91849b4b176067e1434a8c4f94eaf5e56

mvandervoord:
Perfect. Do you just want to do a PR of that and we can merge it in?

Thanks so much for the help!

jlindgren90:
It's included in #454.  Or would you rather have just the first commit in a separate PR?

mvandervoord:
This has been replaced with a cleaner solution by @jlindgren90

PR #454, opened 2019-10-22
Convert RUN_TEST() to a function

jlindgren90:
Converting RUN_TEST() from a macro to a function significantly reduces the size of the compiled binary.  On amd64, the largest test runner in the test suite (testsample_DefaultsThroughCommandLine_runner.o) was reduced from 3.4 kB to 2.4 kB (stripped).

Depends on #453.


jlindgren90:
Parameterized tests are still alive and well, I'm just handling it a little different.  Each parameter set gets wrapped in a tiny void(void) function that then gets handled the same as any other test.  The end result is the same functionality but in a smaller binary.

jlindgren90:
e.g. testparameterized_Runner.c contains:

```
static void runner_args1_test_TheseShouldAllPass(void)
{
    test_TheseShouldAllPass(0);
}
static void runner_args2_test_TheseShouldAllPass(void)
{
    test_TheseShouldAllPass(44);
}
static void runner_args3_test_TheseShouldAllPass(void)
{
    test_TheseShouldAllPass((90)+9);
}
```

And then:

```
  run_test(runner_args1_test_TheseShouldAllPass, "test_TheseShouldAllPass(0)", 78);
  run_test(runner_args2_test_TheseShouldAllPass, "test_TheseShouldAllPass(44)", 78);
  run_test(runner_args3_test_TheseShouldAllPass, "test_TheseShouldAllPass((90)+9)", 78);
```

mvandervoord:
I'm understanding now. I just hadn't dug far enough.
:)

Awesome. I'm definitely all for space savings!

mvandervoord:
Cool. I'm feeling good about these changes. Thanks for all the work @jlindgren90 

@Letme -- I'm pausing here to give you a chance to comment before I merge. I know you're another person who has a lot of valuable thoughts in this area.

PR #455, opened 2019-10-22
Make CMock_Verify() available as UnityRunner_Verify()

jlindgren90:
Having UnityRunner_Verify() accessible from within tests is useful for verifying the call chain halfway through a longer, integration-style test.

This PR is basically equivalent to #376, only rebased on #454 and minimized.

### DIFF
--- a/auto/generate_test_runner.rb
+++ b/auto/generate_test_runner.rb
@@ -231,7 +231,7 @@ class UnityTestRunnerGenerator
 
   def create_mock_management(output, mock_headers)
     output.puts("\n/*=======Mock Management=====*/")
-    output.puts('static void CMock_Init(void)')
+    output.puts('void UnityRunner_Init(void)')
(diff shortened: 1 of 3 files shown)


mvandervoord:
I'm not understanding what this is providing that we didn't already have with `resetTest`?

jlindgren90:
Sometimes it's useful to be able to verify that all the Expects have been fulfilled, without completely resetting the test.  resetTest(), among other things, calls setUp() again, which may clear out important data that's needed in the remainder of the test.

How about just renaming CMock_Verify() to verifyCalls() or something like that?  And then adding prototypes for resetTest() and verifyCalls() in unity.h?  I don't think CMock_Init() and CMock_Destroy() are that useful on their own, so they probably don't need to be made public.

mvandervoord:
Added verifyTest in separate commit

PR #471, opened 2019-12-10
fix float compare

Alanscut:
when `-Wfloat-equal` option is configured, float number compared with `==` will report an error, replace it with `fabs(a-b) < PRECISION` is more safe.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,15 @@ cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
 
 project(unity LANGUAGES C DESCRIPTION "C Unit testing framework.")
 
+set(custom_compiler_flags)
+
+list(APPEND custom_compiler_flags -Wfloat-equal)
+
+# apply custom compiler flags
+foreach(compiler_flag ${custom_compiler_flags})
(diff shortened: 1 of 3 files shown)


LinoMastro:
I don't think these changes are an improvement: you are changing code that is meant to check for exactly 0 and exactly 0.5, both of which can be represented in floating point without errors or approximation.

UNITY_FLOAT_PRECISION by default is relatively large, so this PR significantly changes code behaviour. What's the reason for this, beyond silencing a compiler warning?

And definitely you should not mess with NaN code.

Alanscut:
Many software use the `Unity` test framework, `cJSON` also use it. However, some warnings related to `Unity` will appear during compilation.

LinoMastro:
I understand the desire to avoid compiler warnings but this PR significantly and unnecessarily degrades the precision of UnityPrintFloat for both float and double, and it breaks isnan (after this patch isnan(inf) returns true).

These warnings (float-equal) are overzealous, in some cases a direct comparison of floats with == is perfectly safe and the right thing to do.

There's a reason why both GCC and Clang keep these warnings disabled by default even when both -Wall and -Wextra are enabled: these float-equal warnings have false positives.

If you don't want to see warnings, please don't enable -Wfloat-equal, if you really want/need to enable it, accept that it has false positives and generates some noise.

PR #472, opened 2019-12-10
auto: Only create run_test() function if it's needed.

jlindgren90:
Also fix some whitespace consistency issues.
Use 2 newlines (not 1 or 3) before /*====Heading====*/ comments.

### DIFF
--- a/auto/generate_test_runner.rb
+++ b/auto/generate_test_runner.rb
@@ -92,7 +92,7 @@ class UnityTestRunnerGenerator
       create_suite_setup(output)
       create_suite_teardown(output)
       create_reset(output)
-      create_run_test(output)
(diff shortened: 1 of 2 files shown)


mvandervoord:
I'm not quite following why you would want to run the test runner generator and have it not produce run_test?

jlindgren90:
It happens in code that is tested with multiple build options.  Sometimes all of the tests in a particular file are preprocessed out under a particular build option.

mvandervoord:
And when that happens, was it producing an error? What are we avoiding by removing the function?

jlindgren90:
Not an error, but a warning about an unused static function.  It seemed reasonable to fix the warning at the cost of adding three words ("unless tests.empty?")

PR #473, opened 2019-12-12
Add UNITY_PRINT_TEST_CONTEXT.

jlindgren90:
This option allows you to specify your own function to print additional context
as part of the error message when a test has failed.  It can be useful if you
want to output some specific information about the state of the test at the
point of failure, and UNITY_SET_DETAILS isn't flexible enough for your needs.


mvandervoord:
tweaked and reimplemented.

PR #477, opened 2020-01-05
add support for header files with extensions other than only .h

Tuc-an:
This is an initial step for adding support to CMock to support files with extensions other than only .h and .c. For example this permits easier use with C++ projects that use file extensions like .hh, .hpp, .cc, .cpp, .cxx, etc.


Tuc-an:
@mvandervoord poke

mvandervoord:
Implemented in separate commit to retain ability to separate headers from source.

PR #489, opened 2020-02-14
Feature/report actual code file

jvcdk:
This is an attempt to include the "correct" c-filename in the test output. Before this change, if you used Unity's TEST_-macros in other files than the test-file, you would get wrong filename, but correct line number. Hence you would be looking at the wrong spot to find the output. This patch tries to fix that.

There is one bug; the TEST SUMMARY (probably from Ceedling) still uses the old behavior (i.e. using test-file filename even-though the failing assert is in another (helper) file). I am unsure where the code for that is and how to change it. I was hoping you could spot that bug and fix it (easily)?

Note: This work is based upon the other PR I made; the TEST_PRINTF function.


jvcdk:
Hi @mvandervoord :

I see you have merged #488 ; thank you :-)  I just wanted to check in with you if you had a time to have a look at this as well? 

More specifically, I was wondering whether...
 * the use case is clear?
 * the Ruby/Ceedling "bug" I describe is clear?
 * my assumption that it is easy for you to fix the mentioned bug is reasonable?

If you are busy and just haven't had time to look at it, then no worries. :-) No stress :)  I just wanted to be pro-active and reach out since I realize that *maybe* my PR description could have been better :-D

mvandervoord:
Unity actually started its life with by using `__FILE__` and passing the pointer through all the functions, much like you are doing here. It changed many years ago, for a number of reasons, but mostly these two:

(1) We found that it was most useful to flag the file and line of the **test** that triggered the failure, **not** the line where it was caught. It wasn't nearly as useful to identify the file and line of the failure where it occurred in cmock or unity helper, etc., it was far more useful to know which line of the test it corresponded to. With the emphasis then on passing the correct line number in, the test file was always the file that was being reported.

(2) Those __FILE__ macros all take up space. We support a lot of embedded microprocessors (many of which aren't using particularly awesome compilers) and wasting space on storing information that was almost always the same wasn't worth the memory constraints it added.

So we soon transitioned to the current system. For people who do create helper files with custom assertions (myself included), there is the `UNITY_TEST_ASSERT_` core macros that allow the line to be passed through. This means that you can use all the assertions you want in helper files, and that when they fail, they will report the corresponding line that failed **in the test file** which is most often where it was wanted anyway.

I suspect this isn't the answer you're hoping for. I really do understand where you are coming from. There are times where this functionality would be preferred... but we've not found a low-overhead way of supporting both methods, and the cases where we want to report the test file far outnumber the cases where we want to report anything else.

Does that make sense? (You don't need to agree with me... I just want to make sure I'm being clear about how we got here)

jvcdk:
Hi Mike

Thank you for your response. I do see your point, but I would also like to challenge them a bit :-D

## Add 1)

The current (*) output from Ceedling/Unity for failing tests look like this:
```
-------------------
FAILED TEST SUMMARY
-------------------
[test_devicedetection.c]
  Test: Test_DeviceChanges
  At line (132): "Expected 1 Was 0. Only DIGI0, DIGI1, and TX can change i/o direction."

  Test: Test_SignalNoise
  At line (132): "Expected 1 Was 0. Only DIGI0, DIGI1, and TX can change i/o direction."

  Test: Test_DeviceSignalDropOut
  At line (132): "Expected 1 Was 0. Only DIGI0, DIGI1, and TX can change i/o direction."
```

From this I get the following information:
* Main test file of failing tests is test_devicedetection.c
* Test that failed: Test_DeviceChanges
* Line number: 132.

The line number 132 does not necessary refer to the correct file. If the TEST_ASSERT was made in a helper file, then the line number was in this file. However, only the main test file is given in the output, leading to confusion.

My point is: Assume that this patch was applied (and assuming you helped fix the output of the summary) the correct file would be given, and you would *also* have the name (and test file) of the failing test. So you are not loosing any information :)

(*): When I write "current", it is actually a month old or so; I have not gotten around to upgrade my build environment after you accepted my other pull request, but I will.

## Add 2)

Hm... I see that point. I have not been working on such memory constraint devices, but if you say it is a thing, then it is a thing to take in to consideration.

I think, perhaps, my proposal would have been to let people working with such memory constraint devices divide their test suites up into multiple runners and then only transfer/run a single runner at a time. But I do see that such a way-of-working is more cumbersome.

## Proposal / compromise

I have the idea that the macros can be re-written such that (depending on a compile flag) the file-parameter (to Unity* functions) would be set to either NULL or __FILE__. The code (as today) will then write-out the Unity.TestFile instead.

* => If I can find time to make this change, will you then accept that?
* => And will you put in some work to adapt the Ruby code to output the correct information from the test runners?

I think I would also change functionality such that if NULL is given, then write "N/A" as both filename and line number; outputting a wrong line number is worse than not outputting stuff.

In addition, the line number should be printed next to test case name.

### Example
```
-------------------
FAILED TEST SUMMARY
-------------------
[test_devicedetection.c]
  Test: Test_DeviceChanges (line 94)
  /path/to/gpio_emulation.c (132): "Expected 1 Was 0. Only DIGI0, DIGI1, and TX can change i/o direction."

  Test: Test_SignalNoise (line 144)
  /path/to/gpio_emulation.c (132): "Expected 1 Was 0. Only DIGI0, DIGI1, and TX can change i/o direction."

  Test: Test_DeviceSignalDropOut (line 210)
  /path/to/gpio_emulation.c (132): "Expected 1 Was 0. Only DIGI0, DIGI1, and TX can change i/o direction."
```

I am unsure what output constraints there are in order to support various tools (i.e. VS Code integration), so this is just a uninformed guess ;-)

jvcdk:
@mvandervoord Would you have time to consider my proposal?

mvandervoord:
A different method would definitely be required.

While I DO agree that this can be a useful feature, your discussion is focused on `__FILE__` being the "right" answer most of the time... and this is not true. It's actually uncommon that `__FILE__` is the correct answer when you've configured everything correctly.

Let's say I have a helper file that checks the values in a structure. For simplicity, let's say the structure is something like this:

```
typedef struct POINT_T_
{
    int x;
    int y;
} POINT_T;
```

Now let's say I use this type all over the place and so I want to create a custom assertion into a helper file so it can be used in all my tests. Your first instinct is probably to write it like this:

```
void TEST_ASSERT_EQUAL_POINT_T(POINT_T e, POINT_T a)
{
    TEST_ASSERT_EQUAL_INT(e.x, a.x);
    TEST_ASSERT_EQUAL_INT(e.y, a.y);
 }
```

This would work, but it would return the "wrong" filename with the "right" line number... but here's the thing: it matters far more which TEST failed and which assertion in the test failed. What if `TEST_ASSERT_EQUAL_POINT_T` was called multiple points during a test... reporting the helper file is just not sufficient. So instead, write your helper like this:

```
void Assertion_POINT_T(POINT_T e, POINT_T a, UNITY_LINE_TYPE line, const char* msg)
{
    (void)msg;
    UNITY_TEST_ASSERT_EQUAL_INT(e.x, a.x, line, "X values don't match");
    UNITY_TEST_ASSERT_EQUAL_INT(e.y, a.y, line, "Y values don't match);
 }

// These would be in the header:
void Assertion_POINT_T(POINT_T e, POINT_T a, UNITY_LINE_TYPE line, const char* msg);
#define UNITY_TEST_ASSERT_EQUAL_POINT_T(e,a,l,m) Assertion_POINT_T(e,a,l,m)
#define TEST_ASSERT_EQUAL_POINT_T(e,a) Assertion_POINT_T(e,a,__LINE__,NULL)
#define TEST_ASSERT_EQUAL_POINT_T_MESSAGE(e,a,m) Assertion_POINT_T(e,a,__LINE__,m)
```

So why go through this extra complication?

Two reasons:

(1) Now when a test uses the assertion `TEST_ASSERT_EQUAL_POINT_T`, the actual file and line of the TEST is reported, along with the extra message that lets us know it was the X value. We now have more of the information we need to debug this failure.

(2) The `UNITY_TEST_ASSERT_EQUAL_POINT_T` version can then automatically be detected by CMock and you suddenly are able to mock functions that use POINT_T using your new macro.

Now I realize not all helper files are custom assertions, but most of the same things apply. It's usually more helpful to know where in the test it failed... and the message is available to supply additional details.

Given all that, I agree that there are times where it's just easier to throw an assertion in a helper file and you want the assertion to report where the failure happened. I would like to support that... but I don't want to do it by introducing an extra pointer that has to get passed to every assertion on the list. That's a lot of overhead for something that is almost never used.

The first thing that pops in my head is to add a pair of macros for managing filenames. Maybe something like this:

```
#define UNITY_BEGIN_ASSERTIONS() \
    const char* hold_filename = unity.filename; \
    unity.filename = __FILE__;

#define UNITY_END_ASSERTIONS() \
    unity.filename = hold_filename;
```

Then when you have a helper file, you could simply wrap the contents of functions with these macros and the correct filename (this time meaning the helper) would be reported:

```
void MyHelper(void)
{
    UNITY_BEGIN_ASSERTIONS();

    // does stuff

    TEST_ASSERT_EQUAL_INT(a, b);

     // and more stuff

    UNITY_END_ASSERTIONS();
}
```

It might not be the best idea and I'm definitely open to discussing more. Thanks for getting this conversation moving!

jvcdk:
Hi Mike,

Sorry for the slow reply, but my time has been taken elsewhere.

I see the discussion here actually comes down to what we expect to have in files that are non-strictly-test-files. I created a framework with some device emulation and I see them as part as the test. I included assertions and when they fail, my view is that is the actual place of failure.

I see your view is mainly focused on more simple ASSERT_-extensions, and in these cases, I agree with your view point.

You wrote
> I don't want to do it by introducing an extra pointer that has to get passed to every assertion on the list. That's a lot of overhead for something that is almost never used.

Fair enough. I think you have more experience in what is used when.

I do think, however, the solution with `UNITY_BEGIN_ASSERTIONS` and `UNITY_END_ASSERTIONS` is clunky.

I think I need to leave it here. My work has directed my attention elsewhere and I cannot justify using time on this, sorry. Maybe I will come back to it later.

PR #493, opened 2020-04-09
cmake: update CMake

VShilenkov:
- reorganize CMake's structure
- remove unneeded internal cmake list
- create valid config to allow `find_package(unity)`
- export valid target `unity::framework`
- generate `unity_version.h` header based on version of package in cmake
- include generated header instead of hardcoded 
- use compile options from meson build
- install everything in proper places


michaelbadcrumble:
Hey @VShilenkov. Love the work being done to the CMake support but have you considered the fact that the Meson support may fail because of the version header file? 

Because if someone uses CMake then they are ok, but what about if someone uses Meson then an error would print because of a missing header file.

Is it possible to revert the version changes?

VShilenkov:
Hey @michaelbadcrumble - that was my mistake. I would like to make meson support version and generate that header and then I will be able to support it in cmake.

VShilenkov:
When project version changes - you only need to make version update in root `meson.build` file.
I will put the patch to cmake which will grab this version and use it internally.

VShilenkov:
both builds independently successfull

art-of-dom:
What about people who use neither like Ceedling? Any build system outside these two will break because of lack of having the unity_version.h.

VShilenkov:
Version returned back to main header

michaelbadcrumble:
It’s probably for the best.

PR #497, opened 2020-05-01
More elegant RUN_TEST macro

AlariOis:
Here is a RUN_TEST macro that is equivalent in functionality to the old one but is much more easy to read and understand (in my opinon of course)

The RUN_TEST_AT_LINE macro could be changed to
`#define RUN_TEST_AT_LINE(func, line, ...) UnityDefaultTestRun((func), #func, (line))`
to have an identical output to the old one but that's a few extra parenthesis too many for my taste.

### DIFF
--- a/src/unity_internals.h
+++ b/src/unity_internals.h
@@ -701,11 +701,8 @@ extern const char UnityStrErrShorthand[];
 #endif
 #endif
 #ifdef UNITY_SUPPORT_VARIADIC_MACROS
-#define RUN_TEST(...) UnityDefaultTestRun(RUN_TEST_FIRST(__VA_ARGS__), RUN_TEST_SECOND(__VA_ARGS__))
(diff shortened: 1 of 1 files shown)


mvandervoord:
Reversed because this patch isn't supported in C99

AlariOis:
First of all, sorry for dwelling on this!

When compiling with -std=c99 -pedantic, I indeed got a warning
`ISO C99 requires at least one argument for the "..." in a variadic macro`
I was able to fix that by changing the RUN_TEST macro to
`#define RUN_TEST(...) RUN_TEST_AT_LINE(__VA_ARGS__, __LINE__, throwaway)`

With this change, I get a completely clean build.
I believe this change should make the patch fully supported by c99. Or is there something else I'm still missing?

mvandervoord:
No need to apologize at all! I like the simplification, so long as it's supported! If you'd like to submit as a second PR or reopen this one and tweak it, I'm onboard!

:)

Issue #491, opened 2020-03-27
Missing TEST_ASSERT_NOT_EQUAL_{PTR,STRING,MEMORY} macros


nopeslide:
A lot of dynamic structures use self-referencing when empty.
A macro like `TEST_ASSERT_NOT_EQUAL_PTR` enables simple empty/not-empty checks for such structures.

see linuxs list implementation as an example:
```
/**
 * list_empty - tests whether a list is empty
 * @head: the list to test.
 */
static inline int list_empty(const struct list_head *head)
{
	return READ_ONCE(head->next) == head;
}
```

Further negated macros for strings and memory would enable checking if some memory areas were touched.

laurencedv:
I agree with the last point, being able to assert if some chunk of memory is not at the init value, but whichever it's value would be very useful. :)

AlariOis:
How should these macros act in case of NULL arguments?
FAIL if actual = NULL and expected = NULL (because they are equal)
SUCCEED if actual = NULL and expected = NOT NULL 
SUCCEED if actual = NOT NULL and expected = NULL 

Or would it make more sense to still FAIL in case actual is NULL and expected is not because there is no actual data to compare?

nopeslide:
All of them should behave like the non-inverted versions.
I don't know if the non-inverted versions do any argument checks.

mvandervoord:
The non-inverted variants pass if both are NULL, therefore I would expect this to fail if both are NULL and pass if either is NULL.

nopeslide:
> The non-inverted variants pass if both are NULL, therefore I would expect this to fail if both are NULL and pass if either is NULL.

sounds like

nopeslide:
sorry, I hit the wrong button.

alecoding:
I would expect these macros to do just what they are supposed to do by reading the name:

TEST_ASSERT_NOT_EQUAL_PTR: pass if the pointers differ, no matter if one is NULL or not.  If they are both NULL the macros should report a failure

TEST_ASSERT_NOT_EQUAL_STRING: compare the strings no matter what the pointer is.  If it is NULL too bad.  The tester should add a pointer check before if concerned that it could be not good

TEST_ASSERT_NOT_EQUAL_MEMORY: same as TEST_ASSERT_NOT_EQUAL_STRING

mvandervoord:
[...]

PR #500, opened 2020-05-05
First (far from done) implementation for printf style formatting for …

AlariOis:
I'd like to add printf style formating to all TEST_*_MESSAGE macros as long as  UNITY_INCLUDE_PRINT_FORMATTED  is defined.
I've seen this discussed in some issues and thought to give it a go.

I managed to get a first proof of concept working but as I dont know (yet) the inner working of CMock and Ceedling I'd like some feedback on whether I'm breaking some dependencies. All other suggestions are also welcome of course!

Supported at the moment:
```
TEST_FAIL_MESSAGE
TEST_ASSERT_MESSAGE
TEST_IGNORE_MESSAGE
TEST_MESSAGE
TEST_PASS_MESSAGE
TEST_ASSERT_TRUE_MESSAGE
TEST_ASSERT_UNLESS_MESSAGE
TEST_ASSERT_FALSE_MESSAGE
TEST_ASSERT_NULL_MESSAGE
TEST_ASSERT_NOT_NULL_MESSAGE
TEST_ASSERT_EQUAL_INT[X]_MESSAGE
TEST_ASSERT_EQUAL_UINT[X]_MESSAGE
TEST_ASSERT_EQUAL_HEX[X]_MESSAGE
```
For formating and printing I'm using UnityPrintFVA

The biggest possibly breaking change with CMock and Ceedling is that I had to change the ordering of some function parameters. (UnityFail, UnityIgnore, UnityMessage, UnityAssertEqualNumber)
As long as these tools don't call direclty those functions, everything should be ok.

I used a quite hackish (but short) way to define some functions to support variable argument lists and fallbacks without it.
```c
#ifdef UNITY_INCLUDE_PRINT_FORMATTED
#define VA_LIST_IF_ENABLED , va_list va
#define VA_ARGS_IF_ENABLED , ...
#else
#define VA_LIST_IF_ENABLED
#define VA_ARGS_IF_ENABLED
#endif
```
With this trick its possible to write
```
void UnityFail(const UNITY_LINE_TYPE line, const char* msg VA_ARGS_IF_ENABLED)
```
This expands to
```
void UnityFail(const UNITY_LINE_TYPE line, const char* msg)
```
if UNITY_INCLUDE_PRINT_FORMATTED  is not defined and otherwise to
```
void UnityFail(const UNITY_LINE_TYPE line, const char* msg, ...)
```

Also worth noting is that the _MESSAGE macros are defined like
```c
#define TEST_ASSERT_EQUAL_INT_MESSAGE(expected, actual, ...) 
```
rather than
```c
#define TEST_ASSERT_EQUAL_INT_MESSAGE(expected, actual, format, ...) 
```
This is to support c99, that needs at least one parameter in __VA_ARGS__ list.
With this it's possible to write
```c
TEST_ASSERT_EQUAL_INT_MESSAGE(1, 2, "I cannot believe this failed"); 
```
Without any extra parameters

TEST_PRINTF is now alias to TEST_MESSAGE
And now TEST_PRINTF can also be used without any extra parameters like so:
```c
TEST_MESSAGE("Hello, World!");
TEST_MESSAGE("Hello %s! pi = %f", "World", 3.141592);
```

The ugliest thing about implementing this feature is that all the _MESSAGE macros have to be duplicated. Yikes! Its not possible to use the VA_ARGS_IF_ENABLED  trick with macros sadly.

Again ... all suggestions are welcome! :)

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -561,6 +561,12 @@ void UnityConcludeTest(void)
 }
 
 /*-----------------------------------------------*/
+
+#ifdef UNITY_INCLUDE_PRINT_FORMATTED
+static void UnityPrintFVA(const char* format, va_list va);
+#endif
+
+
 static void UnityAddMsgIfSpecified(const char* msg)
 {
     if (msg)
@@ -584,6 +590,37 @@ static void UnityAddMsgIfSpecified(const char* msg)
         }
 #endif
         UnityPrint(msg);
+    }
+}
+
+static void UnityAddMsgIfSpecified_TEMPORARY_VA_TEST(const char* msg VA_LIST_IF_ENABLED)
+{
+    if (msg)
+    {
+        UnityPrint(UnityStrSpacer);
+
+#ifdef UNITY_PRINT_TEST_CONTEXT
+        UNITY_PRINT_TEST_CONTEXT();
+#endif
+#ifndef UNITY_EXCLUDE_DETAILS
+        if (Unity.CurrentDetail1)
+        {
+            UnityPrint(UnityStrDetail1Name);
+            UnityPrint(Unity.CurrentDetail1);
+            if (Unity.CurrentDetail2)
+            {
+                UnityPrint(UnityStrDetail2Name);
+                UnityPrint(Unity.CurrentDetail2);
+            }
+            UnityPrint(UnityStrSpacer);
+        }
+#endif
+
+#ifdef UNITY_INCLUDE_PRINT_FORMATTED
+        UnityPrintFVA(msg, va);
+#else
+        UnityPrint(msg);
+#endif
     }
 }
 
@@ -705,9 +742,9 @@ void UnityAssertBits(const UNITY_INT mask,
 /*-----------------------------------------------*/
 void UnityAssertEqualNumber(const UNITY_INT expected,
                             const UNITY_INT actual,
-                            const char* msg,
                             const UNITY_LINE_TYPE lineNumber,
-                            const UNITY_DISPLAY_STYLE_T style)
+                            const UNITY_DISPLAY_STYLE_T style,
+                            const char* msg VA_ARGS_IF_ENABLED)
 {
     RETURN_IF_FAIL_OR_IGNORE;
 
@@ -718,7 +755,14 @@ void UnityAssertEqualNumber(const UNITY_INT expected,
         UnityPrintNumberByStyle(expected, style);
         UnityPrint(UnityStrWas);
         UnityPrintNumberByStyle(actual, style);
-        UnityAddMsgIfSpecified(msg);
+        #ifdef UNITY_INCLUDE_PRINT_FORMATTED
+        va_list va;
+        va_start(va, msg);
+        UnityAddMsgIfSpecified_TEMPORARY_VA_TEST(msg, va);
+        va_end(va);
+        #else
+        UnityAddMsgIfSpecified_TEMPORARY_VA_TEST(msg);
+        #endif
         UNITY_FAIL_AND_BAIL;
     }
 }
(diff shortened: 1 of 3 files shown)


mvandervoord:
Heya @AlariOis :

Instead of going through the full list of assertions and creating an alternate version, maybe instead it makes sense to create a single macro which assembles data with sprintf and then it can be dropped into the _MESSAGE portion of any of the existing macros?

AlariOis:
Hey @mvandervoord 
Yes, I have thought of that also. Actually I have used this method in my own projects, but it's a bit clunky to use IMHO. Also I don't like that with this solution the sprintf function runs and wastes CPU cycles every time even if the assertion succeeds.

Do I understand correctly that you proposed something like this:
```c
TEST_ASSERT_TRUE_MESSAGE(false, FORMAT("Hi %s", "there"));
```
or calling the FORMAT macro one level lower like so:
```c
TEST_ASSERT_TRUE_MESSAGE(false, "Hi %s", "there");
```
and in unity.h file
```c
#define TEST_ASSERT_TRUE_MESSAGE(condition, ...) ((condition), __LINE__, FORMAT(__VA_ARGS__))
```

The latter is cleaner to use but then all the _MESSAGE macros still have to have an alternative version. On top of that the sprintf is still running needlessly every time.

AlariOis:
Hi @mvandervoord
I'm almost done with this PR. I think I've covered all the cases now. Have to do some more testing to be sure.
I'd appreciate if you could take a look and test it out :) (no hurry of course)

Maybe there should also be a compiler flag for whether to use UnityPrintFVA or vsprintf for formatting. What do you think?

I know it's a LOT of new macros but I was able to keep the assertion functions in unity.c basically unchanged.

AlariOis:
Resolves #491

Example outputs:
```c
// Pointer comparison
TEST_ASSERT_NOT_EQUAL_PTR(ptr1, ptr1); // FAIL: Expected 0x0040C400 to be not equal to 0x0040C400
TEST_ASSERT_NOT_EQUAL_PTR(NULL, ptr1); // PASS
TEST_ASSERT_NOT_EQUAL_PTR(ptr1, NULL); // PASS
TEST_ASSERT_NOT_EQUAL_PTR(ptr1, ptr2); // PASS
TEST_ASSERT_NOT_EQUAL_PTR(NULL, NULL); // FAIL: Expected 0x00000000 to be not equal to 0x00000000

// String comparison
TEST_ASSERT_NOT_EQUAL_STRING("Foo", "Foo"); // FAIL: Expected 'Foo' to be not equal to 'Foo'
TEST_ASSERT_NOT_EQUAL_STRING(NULL, "Foo"); // PASS
TEST_ASSERT_NOT_EQUAL_STRING("Foo", NULL); // PASS
TEST_ASSERT_NOT_EQUAL_STRING(NULL, NULL); // FAIL: Expected NULL to be not equal to NULL

// String with length comparison
TEST_ASSERT_NOT_EQUAL_STRING_LEN("FooX", "FooY", 3); // FAIL: Expected 'Foo' to be not equal to 'Foo'
TEST_ASSERT_NOT_EQUAL_STRING_LEN(NULL, "Foo", 3); // PASS
TEST_ASSERT_NOT_EQUAL_STRING_LEN("Foo", NULL, 3); // PASS
TEST_ASSERT_NOT_EQUAL_STRING_LEN(NULL, NULL, 3); // FAIL: Expected NULL to be not equal to NULL
TEST_ASSERT_NOT_EQUAL_STRING_LEN("Foo", "Foo", 0); // FAIL: Expected '' to be not equal to ''
TEST_ASSERT_NOT_EQUAL_STRING_LEN(NULL, NULL, 0); // FAIL: Expected NULL to be not equal to NULL
TEST_ASSERT_NOT_EQUAL_STRING_LEN("Foo", NULL, 0); // PASS
TEST_ASSERT_NOT_EQUAL_STRING_LEN(NULL, "Foo", 0); // PASS

// Memory comparison
TEST_ASSERT_NOT_EQUAL_MEMORY("Foo", "Foo", 3); // FAIL: Expected 3 bytes of memory not to be equal
TEST_ASSERT_NOT_EQUAL_MEMORY(NULL, "Foo", 3); // PASS
EST_ASSERT_NOT_EQUAL_MEMORY("Foo", NULL, 3); // PASS
TEST_ASSERT_NOT_EQUAL_MEMORY(NULL, NULL, 3); // FAIL: Expected NULL to be not equal to NULL
TEST_ASSERT_NOT_EQUAL_MEMORY("Foo", "Foo", 0); // FAIL: You Asked Me To Compare Nothing, Which Was Pointless.
```

To keep from making 3 totally new assertion functions for these cases I modified UnityAssertEqualString, UnityAssertEqualStringLen, UnityAssertEqualMemory functions by adding a new `const UNITY_UINT not` parameter to them. If this is 0 then these functions act as before and if it is 1 then the results are inversed.

I guess I should add new unity test cases for these new assertion macros.
Also I think ideally the test cases should test _MESSAGE macros compiling with and without  UNITY_INCLUDE_PRINT_FORMATTED defined to test __VA_ARGS__ stuff

PR #522, opened 2020-09-07
Add guard TEST on test module template

JuPrgn:
With a test file guarded we can include this file on IDE project (MPLAB X in my case) and compile without excluding test files.
Excluding test files on MPLAB X disable autocompletion and function navigation.

Please note that I am not sure everyone define TEST on their projects and if this should be an option instead of the default template.


mvandervoord:
I feel like TEST should be the standard, so I'm all for it.

PR #526, opened 2020-10-08
Support for compiling with addons enabled

### LINKED ISSUE #521: Can't compile tests with unity_fixture.h and current CMakeList.txt
Hi,
I'm trying to compile unity tests with the currently provided CMakeLists.txt and while it works when using unity.h if I use Grenning's approach using unity_fixture.h I get an error without modifying the list like this:

```
target_sources(${PROJECT_NAME}
    PRIVATE
        src/unity.c
        extras/fixture/src/unity_fixture.c
        extras/memory/src/unity_memory.c
)

target_include_directories(${PROJECT_NAME}
    PUBLIC
        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/extras/fixture/src>
        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/extras/memory/src>
        $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}>
)
```

When I add the sources and the include directories I can compile using the two approaches. Would there be any drawbacks for anyone not using the fixture? Any chance of making a PR for this? Am I doing it completely wrong?

Thanks a lot in advance.

gmircea:
Unity provides extra addons, fixtures and memory, that are not
included in the build by default.

This commit adds support for enabling and installing Unity with the
addons included:

    cmake -DWITH_ADDONS=ON [path to top level Unity directory]

If the above option is not provided, the addons are not included in
the build.

Fixes #521

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,8 @@ project(unity
     DESCRIPTION "C Unit testing framework."
 )
 
+option(WITH_ADDONS "Compile with support for addons: fixtures and memory" OFF)
+
 # Main target ------------------------------------------------------------------
 add_library(${PROJECT_NAME} STATIC)
 add_library(${PROJECT_NAME}::framework ALIAS ${PROJECT_NAME})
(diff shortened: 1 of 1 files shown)


mvandervoord:
Thanks for the help! I'm going with @carsonalh's similar submission because it's a bit more complete, but I really appreciate your time and work here!

PR #528, opened 2020-11-25
set lower bound for delta in float/double equal comparison

shintarokkk:
In equality check of float/double, "_UNITY\_<FLOAT/DOUBLE>\_PRECISION * expected_" is used as "_delta_" of _TEST\_ASSERT\_<FLOAT/DOUBLE>\_WITHIN\_MESSAGE(delta, expected, actual, message)_.

That can cause too exact comparison when "_expexted_" is equal to or near zero.

Therefore, I think it is better to set a lower bound for "_delta_" when it will be smaller than precision. I made a pull request that set "_UNITY\_<FLOAT/DOUBLE>\_PRECISION_" as the lower bound in such case.

For example, the following tests all fail in current master branch. My pull request can modify these tests to act in the way it's _supposed_ (at least I suppose) to be.

<details>
<summary>test code</summary>
<pre>
<code>
#include "unity.h"

void setUp(void)
{
}

void tearDown(void)
{
}

void test_FloatNearZero_SupposedToPass(void)
{
  TEST_ASSERT_EQUAL_FLOAT(0.0, 0.0000000000000000000000001f);
}

void test_FloatNearZero_SupposedToFail(void)
{
  TEST_ASSERT_EQUAL_FLOAT(0.0, 0.0001f);
}

void test_DoubleNearZero_SupposedToPass(void)
{
  TEST_ASSERT_EQUAL_DOUBLE(0.0, 1e-300);
}

void test_DoubleNearZero_SupposedToFail(void)
{
  TEST_ASSERT_EQUAL_DOUBLE(0.0, 1e-11);
}

void test_FloatArrayNearZero_SupposedToPass(void)
{
  float exact[2] = {0.0, 0.0};
  float input[2] = {0.000000000001f, -0.000000000001f};
  TEST_ASSERT_EQUAL_FLOAT_ARRAY(exact, input, 2);
}

void test_FloatArrayNearZero_SupposedToFail(void)
{
  float exact[2] = {0.0, 0.0};
  float input[2] = {0.0001f, -0.0001f};
  TEST_ASSERT_EQUAL_FLOAT_ARRAY(exact, input, 2);
}

void test_DoubleArrayNearZero_SupposedToPass(void)
{
  double exact[2] = {0.0, 0.0};
  double input[2] = {1e-100, -1e-100};
  TEST_ASSERT_EQUAL_DOUBLE_ARRAY(exact, input, 2);
}

void test_DoubleArrayNearZero_SupposedToFail(void)
{
  double exact[2] = {0.0, 0.0};
  double input[2] = {1e-10, -1e-10};
  TEST_ASSERT_EQUAL_DOUBLE_ARRAY(exact, input, 2);
}

int main(void)
{
  UNITY_BEGIN();
  RUN_TEST(test_FloatNearZero_SupposedToPass);
  RUN_TEST(test_FloatNearZero_SupposedToFail);
  RUN_TEST(test_DoubleNearZero_SupposedToPass);
  RUN_TEST(test_DoubleNearZero_SupposedToFail);
  RUN_TEST(test_FloatArrayNearZero_SupposedToPass);
  RUN_TEST(test_FloatArrayNearZero_SupposedToFail);
  RUN_TEST(test_DoubleArrayNearZero_SupposedToPass);
  RUN_TEST(test_DoubleArrayNearZero_SupposedToFail);
  return (UNITY_END());
}
</code>
</pre>
</details>

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -912,6 +912,7 @@ void UnityAssertEqualFloatArray(UNITY_PTR_ATTRIBUTE const UNITY_FLOAT* expected,
     UNITY_UINT32 elements = num_elements;
     UNITY_PTR_ATTRIBUTE const UNITY_FLOAT* ptr_expected = expected;
     UNITY_PTR_ATTRIBUTE const UNITY_FLOAT* ptr_actual = actual;
+    UNITY_FLOAT delta;
 
     RETURN_IF_FAIL_OR_IGNORE;
 
@@ -932,7 +933,12 @@ void UnityAssertEqualFloatArray(UNITY_PTR_ATTRIBUTE const UNITY_FLOAT* expected,
 
     while (elements--)
     {
-        if (!UnityFloatsWithin(*ptr_expected * UNITY_FLOAT_PRECISION, *ptr_expected, *ptr_actual))
+        delta = *ptr_expected * UNITY_FLOAT_PRECISION;
+        if (delta < UNITY_FLOAT_PRECISION && delta > -UNITY_FLOAT_PRECISION)
+        {
+            delta = UNITY_FLOAT_PRECISION;
+        }
+        if (!UnityFloatsWithin(delta, *ptr_expected, *ptr_actual))
         {
             UnityTestResultsFailBegin(lineNumber);
             UnityPrint(UnityStrElement);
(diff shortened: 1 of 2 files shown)


shintarokkk:
I read the [document](https://github.com/ThrowTheSwitch/Unity/blob/386c540510929c0cf6f7b8d04cb4af40754bb822/docs/UnityAssertionsReference.md) again and noticed the behavior of delta around zero was intentional. Sorry for that.

> So what happens when it's zero? Zero - even more than other floating point values - can be represented many different ways. It doesn't matter if you have 0 x 20 or 0 x 263.It's still zero, right? Luckily, if you subtract these values from each other, they will always produce a difference of zero, which will still fall between 0 plus or minus a delta of 0. So it still works!

We can manage this if we use _TEST\_ASSERT\_DOUBLE\_WITHIN_ when we deal with scalar, so I will close this pull request.

However when it comes to vectors and matrices, Unity does not have macros to solve that issue by default. And that is why I made this pull request. I will make another pull request for _TEST\_ASSERT\_<DOUBLE/FLOAT>\_ARRAY\_WITHIN_.

mvandervoord:
@shintarokkk -- I wouldn't be against the idea of making it configurable as to which option to use. You're not the first person to wish they worked the other way.

PR #537, opened 2021-02-04
New feature/branch: Create tests that can run without having to register them.

carsonalh:
**IMPORTANT**: I don't want this merged with master yet, so _please_ don't do that. I would much rather have another branch created where I can progressively work on this feature over time.

This commit marks the start of a new feature that I'm trying to
implement to get unity to work without having to register each and every
test in order to run it.  The idea is that tests are run as blocks in a
larger function instead of being functions themselves.  This way, there
is no need to register each function, as they will run in the same way
that normal code inside a function runs; because that's what it is.

The way it works is using the `TEST` macro that I've defined.  This
calls an internal function I've called `unity_register_test_` (naming is
flexible), which changes Unity's internal global state to say that the
following code is a part of the next test.  Another idea I have in mind
(but not yet implemented), is to have a function that, like
`unity_register_test_`, is automatically inserted with the `TEST` macro,
but instead looks at the global state _after_ a test has run to report
its success / failure / ignore.  Since the block after the test macro is
where the code lies, this function would have to account for the
_previous_ test.  To account for the case of the last test in the suite,
that could be done in main, after `unity_suite_func_` is called.

For now, however, this is still basically in its concept phase, and any
further development would be appreciated!


Crazor:
This looks like a promising idea. Are you still working on this branch?

carsonalh:
I can still work on it if you want. I thought it wasn't going to be accepted. I'd be more than happy to stay working on it again.

Crazor:
Even in this was not accepted, I would want to use something like this. As it is now, Unity does not follow the DRY principle, which is unnecessarily verbose and also a cause for errors if I'm not careful. I was actually thinking of creating a fork to fix this, but then I came across your effort.

mvandervoord:
I'd love to see continued work on this (it's nice to have an alternative to using a test runner generated or having to do it manually). I'm closing this PR, though, as there has been no progress on a long time and the branch can continue to exist without it.

Issue #510, opened 2020-06-26
Warnings, during the compilation.


vladipus:
This is kind of irritation and halts the build when using `-Werror`:
```
unity.c(762,28): warning: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Wextra-semi-stmt]
[build]         UNITY_FAIL_AND_BAIL;
```

vladipus:
And also these on the generated code:
```
int GlobalExpectCount;
[build]     ^
[build] test\test_***_Runner.c(17,1): note: declare 'static' if the variable is not intended to be used outside of this translation unit
```

PR #544, opened 2021-02-27
Bugfix/fix issue 510

farrrb:
Fix first part of #510 in `unity.c` . 

**Note:**  This does ___not___ fix the `int GlobalExpectCount;` error  @vladipus is seeing. 

There is also an error in my CI runs (but not in the official ones), that will get fixed by this PR (see https://github.com/ThrowTheSwitch/Unity/commit/fa32e530ba1aecc37113067714d29459d12bbdec):
image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,7 @@ target_compile_options(${PROJECT_NAME}
                                  -Wcast-qual
                                  -Wconversion
                                  -Wexit-time-destructors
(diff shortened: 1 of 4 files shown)


farrrb:
Just for the sake of completeness, here is a screenshot of the internal run from github workflow... I cannot comprehend what was going on there and why it was only on my side (my fork) of the repository...
image

PR #545, opened 2021-03-09
Test define customizable in module_generator

MrRedKite:
Hello,
I had the problem, that the recently added "#ifdef TEST" in the testfile didn't work, because I used the define "TEST" for other purposes. This is why I made the define customizable via the project.yml in ceedling. With this, you can use your own define (in my case "CEEDLING_TEST").

I hope its useful :)

Related to Ceedling pull request: [ThrowTheSwitch/Ceedling#570](https://github.com/ThrowTheSwitch/Ceedling/pull/570)

### DIFF
--- a/auto/generate_module.rb
+++ b/auto/generate_module.rb
@@ -13,7 +13,7 @@ require 'fileutils'
 require 'pathname'
 
 # TEMPLATE_TST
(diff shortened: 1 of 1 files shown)


mvandervoord:
Merge to PR #656

PR #553, opened 2021-04-07
Fix unity_test_summary ruby script with parameterized tests

jannisbaudisch:
When you passed a string with a colon (":") to a test via test
parameterization, the unity_test_summary.rb script was no longer
able to handle its output.

EXAMPLE:
```
TEST_CASE("::1")
void test_some_parameterized_ipv6_address(const char *address)
{
	TEST_FAIL();
}
```

The reason for this is, that a simple split at colon command was used.
Changed this to a more complex regex.

PLEASE NOTE: I have no clue, how to write clean ruby code. I bet this
can be written much prettier. Feel free to change it, but at least the
concept is working.


mvandervoord:
Merge into PR #656

PR #555, opened 2021-04-16
Ignore space around parameter in TEST_CASE()

erijo:
This makes it possible to use defines that expand to something that
includes space, e.g. TEST_CASE(true).


erijo:
Anything I can do to help push this PR forward?

PR #556, opened 2021-04-16
Add support for TEST_RANGE with exclusive end

erijo:
If the range is <start, end, step> instead of [start, end, step], the
end value will not be included in the range.

This can be useful if you have a define that defines e.g. the size of
something and you want to use this define as the end value. As the
pre-processor doesn't evalutate expressions (unless you do some macro
magic) you can't specify the range as [0, MY_SIZE - 1, 1]. With this
change you can then instead give the range <0, MY_SIZE, 1>.

### DIFF
--- a/auto/generate_test_runner.rb
+++ b/auto/generate_test_runner.rb
@@ -148,12 +148,13 @@ class UnityTestRunnerGenerator
           end
 
           # RANGE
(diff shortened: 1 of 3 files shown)


erijo:
What's your thought on this PR @mvandervoord? Do you think it can be accepted or would you rather see it done in some other way?

mvandervoord:
I like this. If it's easy enough, can you throw in a test case or two to show it works (and as an example of what's possible)

erijo:
I've an attempt at a test now @mvandervoord.

PR #557, opened 2021-04-17
Fix failing unit tests

erijo:
Fix warnings/error in unittests.

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -445,7 +445,7 @@ void UnityPrintFloat(const UNITY_DOUBLE input_number)
 
         /* build up buffer in reverse order */
         digits = 0;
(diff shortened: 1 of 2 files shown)


src/unity.c
```diff
@@ -445,7 +445,7 @@ void UnityPrintFloat(const UNITY_DOUBLE input_number)
 
         /* build up buffer in reverse order */
         digits = 0;
-        while ((n != 0) || (digits < (decimals + 1)))
```
KalleDK, line 448: Do you need this check at all ? As digits is always zero and decimals can't be less than zero?

erijo: digits is incremented in the loop.

KalleDK: Argh my bad for missing that! Just smile and wave boys :D

PR #578, opened 2021-09-18
'Reformat' branch

Kochise:
Indent changed to TAB (more flexible and more compact)
Macro renamed (better visual context)
Macro moved (grouping message and type)
Added embedded framework (specific code)
Added logging (target to configure)


Kochise:
Ok, might be ok, do whatever with it.

These changes have been around on my hd since 2013/2014 (in C++ though), converted them into C in 2017, made some changes in early 2019 without pushing them up, lagged behind for another couple of years, now I synced them with the recent changes and it's in the wild.

I may add a few more tiny enhancements in the following days.

mvandervoord:
Hm. 

I appreciate all the work you've put into this. I suspect there is some good stuff in here, but I'll have to sift through it. 

We have a coding standard which your recent changes conflict with, so we won't be accepting changes that conflict with that. 

We have thousands of active users, so won't be merging naming changes that aren't there for a good reason.

I see you've added a different fixture. It seems to mostly reproduce the fixture that exists in options already, but with different function names... I'll dig into that in more detail to see if there are some gems there.

Thanks for the snapshot. I know how tools evolve to meet your own needs and I appreciate your sharing where you've ended up. I appreciate the opportunity to compare your snapshot to our own evolution and borrow some of it!

Kochise:
No problem, I've added `unity_legacy.h` to avoid breaking existing code.

Yeah, as I said, do whatever you want, it's been a pleasure to work on you code, so clean and concise. It's amazing how much it is reliable and well known in the industry.

The PLAN/FIXT/CASE/STEP was for a complete framework to structure unit testing and automate their execution, collect results and conclude in a hierarchical manner. Syntax might be weird at first, but it's pretty effective. No threading support, though.

Regards.

Kochise:
Btw, feel free to ask me anything about this code.

The naming changes were mostly for visual context (double underscores to help spot the asserts' types) and coherence (EQUAL first, EQUAL_ARRAY then EQUAL_EACH, all starting with EQUAL, same for WITHIN).

I've also added some missing macros, but not all (NOT_EQUAL_BITx for instance).

mvandervoord:
Likely will cherry-pick ideas from this, but these fixes are far too broad and we won't be merging to the main project.

PR #598, opened 2022-02-16
Parse output color & time support with custom test suite name

AJIOB:
1) Adding support for colorful Unity output (with defined `UNITY_OUTPUT_COLOR` value)
2) Adding the possibility for defining custom test suite name for generated JUnit report.
3) Adding support for non-ASCII characters for correct output with other languages.
4) Adding support for test execution time parsing

### DIFF
--- a/auto/parse_output.rb
+++ b/auto/parse_output.rb
@@ -57,6 +57,10 @@ class ParseOutput
     puts 'Real test suite name will be \'' + @real_test_suite_name + '\''
   end
 
+  def xml_encode_s(s)
+    return s.encode(:xml => :attr)
+  end
+
   # If write our output to XML
   def write_xml_output
     output = File.open('report.xml', 'w')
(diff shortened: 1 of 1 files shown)


mvandervoord:
Hi Alex:

The title of this post discusses color, but your changes all seem to be towards reporting test time.

The time output is an optional output with Unity. Do these changes continue to work properly if the time isn't included?

Mark

AJIOB:
Hi Mark

As you can see, Unity-generated outputs will generate correct XML files. If you pass the `-suiteHelloWorld` flag, the default `Unity` test suite name will be replaced with `HelloWorld` in XML output.

File with 2 different outputs in 4 different forms (with/without color & with/without time) is available [here](https://github.com/ThrowTheSwitch/Unity/files/8153513/UnityDifferentOutputs.zip).

With regards,
Alex

PR #599, opened 2022-02-16
Adding test matrix feature

AJIOB:
As an analog for `TEST_CASE(...)` & `TEST_RANGE(...)`, you can use `TEST_MATRIX(...)` for defining custom test cases.

`TEST_MATRIX` macro parsing can be enabled together with `TEST_RANGE`.

Supports almost all found C & C++ literals & expressions (expect array addressing with `[]`, such as `my_arr[26]`). Also supports multiline matrixes, ranges & cases.

Examples of some tests with that usage:
```c
TEST_MATRIX([ "World", "More \"words", "than ]", "one)", NULL ], [ 5, 8, 4, 32 ])
TEST_MATRIX([
        NULL, "name",
        "name1", "2name", "33name"
    ], [4])
void test_nothing(const char *s, uint32_t b)
{
    TEST_ASSERT_NOT_NULL_MESSAGE(s, "Bad string");
    TEST_ASSERT_NOT_EQUAL_UINT32_MESSAGE(0, b, "Number is zero");
}
```

### DIFF
--- a/auto/generate_test_runner.rb
+++ b/auto/generate_test_runner.rb
@@ -167,7 +167,7 @@ class UnityTestRunnerGenerator
         ## Args with spaces (not strings or char sequences)
         ## String concatenations: "Hello" "World"
 
-        one_number_or_text_regex_string = /(?:[^"',\s\]](?:[^"',\]]*[^"',\s\]])*|(?:(?:\+\-)\s*)?\d+(?:'\d+)*)/.source
+        one_number_or_text_regex_string = /(?:[^"',\s\]](?:[^"',\]][^"',\s\]]?)*|(?:(?:\+\-)\s*)?\d+(?:'\d+)*)/.source
         one_arg_regex_string = /(?:(?:(?:#{one_number_or_text_regex_string}\s*|)"(?:[^"\\]|\\.)*")+|(?:#{one_number_or_text_regex_string}\s*|)'(?:[^'\\]|\\.)*'|(?:#{one_number_or_text_regex_string}|))\s*/.source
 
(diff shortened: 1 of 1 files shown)


AJIOB:
Hello @mvandervoord,

Have I lost some requirements for merging that merge?

With regards,
Alex

mvandervoord:
Hi Alex:

I'm trying to figure out a good way to test all of this. It's going to require tests to make sure the script itself performs the job assigned, as well as that both the code using the new notation and the generated code both actually perform their job and compile without error. 

Also, we're going to want to make sure the documentation for the new feature gets added.

These things don't all have to be on you, necessarily, but they're important for a significant feature like this.

AJIOB:
Hi Mark

As you can see, Unity-generated outputs will generate correct XML files. If you pass the `-suiteHelloWorld` flag, the default `Unity` test suite name will be replaced with `HelloWorld` in XML output.

File with 2 different outputs in 4 different forms (with/without color & with/without time) is available [here](https://github.com/ThrowTheSwitch/Unity/files/8153513/UnityDifferentOutputs.zip).

With regards,
Alex

jonhenneberg:
@AJIOB and @mvandervoord Was this abandoned for some other/better solution ?

I'm interested in the feature, and was considering adding something to our local fork of Unity, but would rather add it to the upstream project, so if it wasn't dropped due to some other design or an incompatibility, I was contemplating brushing this PR off, getting it updated and looking at what could be done to add tests to it.

mvandervoord:
I just haven't had time to get back on this, unfortunately. :(

mvandervoord:
@jonhenneberg -- it would not be a wasted effort. In fact, it would be much appreciated. :)

PR #601, opened 2022-02-28
unity_to_junit.py can be imported as Python module now

AJIOB:
Based on https://docs.python.org/3/tutorial/modules.html#packages


AJIOB:
Hello @mvandervoord 

What about that fix merging?

Maybe I miss something more?

With regards,
Alex

joukewitteveen:
This is not good as it implicitly creates a Python package called `auto`. Where was this needed? You could always import the file as a module when it was on the Python search path. For interactive sessions, the Python search path includes the working directory, meaning that when you start Python from `auto/`, you could `import unity_to_junit`.

I suggest reverting the commit from this merge request.

Letme:
On a side note, we created a separate python package for exactly this conversion (because there is no unity_to_junit.py in repo) https://github.com/melexis/unity2junit

Issue #568, opened 2021-07-07
Compilation error with Microsoft Visual Studio Community 2019


sdelisle25:
Dear,

I try to compile Unity master **branch** (2f7406572e6f11212536c99a839e1bf512fe3fb6) by using Visual Studio Community 2019 (Plateform Toolset: Visual Studio 2019 (v142)), 

_Unity\src\unity_internals.h(57,1): fatal error C1083: Cannot open include file: 'stdnoreturn.h': No such file or directory_

By using  Plateform Toolset: Visual Studio 2017 (v141), the compilation is ok (__STDC_VERSION__  < 201112L).

I tried to use Plateform Toolset: LLVM (clang-cl) provided by Visual Studio 2019, by the compilation fail on different errors.

Besr regards
Sylvain Delisle

AJIOB:
Hello @sdelisle25 

Looks like current master branch has fixes for your issue.

Maybe you can check & close issue.

With regards,
Alex

PR #604, opened 2022-03-04
Fixing noreturn usage on old Windows SDKs with new MSVC compiler

AJIOB:
As you can see [here](https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/), Microsoft implements `stdnoreturn.h` header in the latest SDKs only (Windows 10 SDK 2104+).

But if you are using the latest MSVC with C11 support with old Windows 10 SDK, you cannot compile your code.

That fix should fix working with all Windows SDK versions.

Commit was tested locally with Windows SDK 10.0.20348 (2104), 10.0.19041 (2004), 10.0.16299 (1709) & MSVC 14.30.

P.S. As you can see in [Microsoft docs](https://docs.microsoft.com/en-us/cpp/c-language/noreturn?view=msvc-170) & [cppreference](https://en.cppreference.com/w/c/language/_Noreturn), `stdnoreturn.h` can be optionally implemented and also [have some problems with GCC](https://stackoverflow.com/q/28552467/6818663).
P.P.S. Example of using that macro can be found [here](https://github.com/microsoft/DirectXShaderCompiler/blob/407f8ebf8df59d9c848a0da39c1e408596891495/tools/clang/unittests/HLSL/ExecutionTest.cpp#L1234).

### DIFF
--- a/src/unity_internals.h
+++ b/src/unity_internals.h
@@ -52,8 +52,28 @@
       #define UNITY_NORETURN [[ noreturn ]]
     #endif
   #elif defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
-    #include <stdnoreturn.h>
-    #define UNITY_NORETURN noreturn
+    #if defined(_WIN32) && defined(_MSC_VER)
+      /* We are using MSVC compiler on Windows platform. */
+      /* Not all Windows SDKs supports <stdnoreturn.h>, but compiler can support C11: */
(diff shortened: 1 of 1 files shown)


AJIOB:
@mvandervoord 
Hello Mark

Looks like it fixes an issue #568 too.

What did I miss for accepting that PR?

With regards,
Alex

PR #608, opened 2022-04-22
List Unity framework in PlatformIO Registry

ivankravets:
In continuation to https://github.com/ThrowTheSwitch/Unity/pull/607

This PR allows PlatformIO users to use the original version of Unity. The v2.5.2 with #607 fix is already published ( https://registry.platformio.org/libraries/throwtheswitch/Unity ). We will use the official version instead of our internal in the upcoming PlatformIO Core 6.0.

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,0 +1,15 @@
+{
+  "name": "Unity",
+  "version": "2.5.2",
+  "keywords": "unittest, tdd, unit, test",
+  "description": "Simple Unit Testing for C",
+  "homepage": "http://www.throwtheswitch.org/unity",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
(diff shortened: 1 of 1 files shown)


ivankravets:
@mvandervoord thanks for merging both PRs! 🙏 

Sorry for the extra work for you regarding "libraru.json". Next time before making a new Git TAG, you will need to increment the `version` field in the manifest.

P.S: Do you plan to release 2.5.3 or 2.6.0? A lot of changes since the 2.5.2 version.

mvandervoord:
No problem. I'll just add it to my `release` script. :)

Yes. I'm working towards a new release right now, actually. I'm just working through the backlog of PR's and issues to see if there are noteworthy things that should be included before pulling the plug.

:) 

Thanks for the help and the exposure! PlatformIO is a great tool!

PR #618, opened 2022-06-19
Updating the Meson build scripts

troglobyte-coder:
I'm updating the Meson build scripts to 62 for the latest bug fixes and using the new wrap-file feature.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,14 +1,12 @@
 #
-# build script written by : Michael Brockus.
+# build script written by : Michael Gene Brockus.
 # github repo author: Mike Karlesky, Mark VanderVoord, Greg Williams.
 #
 # license: MIT
 #
 project('unity', 'c',
-    license: 'MIT',
-    meson_version: '>=0.53.0',
-    default_options: ['werror=true', 'c_std=c11']
-)
+    meson_version: '>=0.62.0',
+    default_options: ['werror=true', 'c_std=c11'])
 
 subdir('src')
 unity_dep = declare_dependency(link_with: unity_lib, include_directories: unity_dir)
--- a/examples/example_4/meson.build
+++ b/examples/example_4/meson.build
@@ -6,7 +6,7 @@
 #
 project('example-4', 'c')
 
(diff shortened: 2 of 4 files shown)


meson.build
```diff
@@ -1,14 +1,12 @@
 #
-# build script written by : Michael Brockus.
+# build script written by : Michael Gene Brockus.
 # github repo author: Mike Karlesky, Mark VanderVoord, Greg Williams.
 #
 # license: MIT
 #
 project('unity', 'c',
-    license: 'MIT',
-    meson_version: '>=0.53.0',
```
eli-schwartz, line 9: Hmm, what's the reason for bumping the minimum version of Meson here?

Also, why remove the license info? Very confusing.

troglobyte-coder: License info just seems like extra metadata. Meson version was bumped to ensure we can use wrap-file [provides](https://mesonbuild.com/Wrap-dependency-system-manual.html#provide-section) in the sample and more importantly so the users can use it with that new feature, plus to make use of any bug fixes.

eli-schwartz: > License info just seems like extra metadata.

It's legally useful. https://en.wikipedia.org/wiki/Software_supply_chain

> Meson version was bumped to ensure we can use wrap-file [provides](https://mesonbuild.com/Wrap-dependency-system-manual.html#provide-section) in the sample and more importantly so the users can use it with that new feature, plus to make use of any bug fixes.

This is false, you may have intended to do it for that reason, but bumping the version did not accomplish any such thing.
- the provide section is fully backwards compatible, only the example needs to be built with newer Meson (but the example does not specify a minimum!)
- as the documentation you linked to states, you only need 0.55.0, not 0.62.0
- "to make use of any bug fixes" is not the decision of this project to make. Users who want bug fixes in Meson shall upgrade Meson themselves, this project doesn't need to impose that on them. Unless it's a bugfix for this project, but based on your wording, it's not clear that there are any bug fixes you are considering?

troglobyte-coder:
Fair enough, I’ll fix this after it cools down.

PR #623, opened 2022-06-21
Adding dependabot and moved .files

troglobyte-coder:
Added a dependabot file to automatically fetch updates for GitHub action files, also moved .files into the `.github` directory.


Letme:
Care to explain why .gitignore is moved into github directory? .gitignore is used by git, has nothing to do with github, and also .gitattributes. So makes no sense to move them even further in some subdirectory

PR #655, opened 2023-02-01
Declare functions static in the TEST() macro

johannesacco:
This makes it possible to use "-Wunused-function" and get a compiler warning when a test have been declared but not added to TEST_GROUP_RUNNER().


johannesacco:
I see now that in the automated tests the test runner is declared in another file which does not work with this change. I will close this.

PR #657, opened 2023-02-09
Enable the collection of test of a directory tree

flieger19:


mvandervoord:
I'm not clear on what this update is building? The `main` test runner means what?

flieger19:
AFAIK the `generate_test_runner.rb` script can only generate a runner for a single source file with tests. But not for several source files with tests a single common runner.
Accordingly, I have extended the script so that it can also iterate through a directory tree and scan all matching source files there for tests. These are then collected in a common test runner (`main` runner).

tango-taylor:
This would be super useful, I was looking for this exact functionality. I have some tests in C and some in C++ and it seems the easiest way to run them both would be to link one test runner against them.

tango-taylor (commented):
@flieger19 I gave this a try, I think it's a great start. To be complete I think it should have the following: 

* The usage/help text needs to be updated to let the user know it takes directories as input instead of just files. 
* Add some unit tests. This gives usage examples "for free" and builds confidence. The last thing we want is the runner generator to drop tests. Looks like `test_generate_test_runner.rb` would be a good place to start, I'm sure @mvandervoord would know more (I'm not a dev here). 
* To keep compatibility with how the script handles files right now, it'd be better if it placed all the tests into *one* runner file. e.g. I tried it with `test.c` and `test_cpp.cpp` and it gave me three output files: `test_Runner.c`, `test_cpp_Runner.cpp`, `testAll_Runner.c`. It would be better if it gave me just `testAll_Runner.c` (or make the name of the output file mandatory so it doesn't have to make up a name). This might also make it easier to integrate with `test_generate_test_runner.rb` for testing.

PR #546, opened 2021-03-14
Meson updates

optorres:
This updates the meson build scripts to bring them more-or-less up to parity with the CMake script and adds a meson build script to one of the project examples.

Build script additions:
 * Library version is now retrieved from `unity.h` instead of being left as undefined.
 * Extensions (`fixture` and `memory`) are now supported.
 * The library, headers, and package configuration files (using `pkg-config` instead of CMake's config format) are now installed with `ninja install` / `meson compile install`.

PR #658, opened 2023-02-13
Meson update

amcnulty-fermat:
This PR is heavily based on existing work by Owen Torres (@optorres) in #546. I aim to get his excellent work merged. 

As stated by @optorres in #546 : 

> This updates the meson build scripts to bring them more-or-less up to parity with the CMake script and adds a meson build script to one of the project examples.
> 
> Build script additions:
> 
> Library version is now retrieved from `unity.h` instead of being left as undefined.
> Extensions (`fixture` and `memory`) are now supported.
> The library, headers, and package configuration files (using `pkg-config` instead of CMake's config format) are now installed with `ninja install` / `meson compile install`.

My additions to his work are mainly around hygiene for Meson.  I try to use the cross-platform `/` operator to join paths, which was introduced in version '0.49.0' and use `meson.project_source_root()` which is introduced in Meson version `0.56.0` in place of the deprecated function `meson.source_root()`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -5,21 +5,22 @@
 # license: MIT
 #
 project('unity', 'c',
-    license: 'MIT',
-    meson_version: '>=0.37.0',
-    default_options: ['werror=true', 'c_std=c11'])
+  license: 'MIT',
+  # `meson.project_source_root()` introduced in 0.56.0
+  meson_version: '>=0.56.0',
+  default_options: [
+    'werror=true',
+    'c_std=c11'
+  ]
+)
 
 subdir('src')
 unity_dep = declare_dependency(link_with: unity_lib, include_directories: unity_dir)
 
-
-# Get the generate_test_runner script relative to itself or the parent project if it is being used as a subproject
-# NOTE: This could be (and probably is) a complete hack - but I haven't yet been able to find a better way....
-if meson.is_subproject()
-gen_test_runner_path = find_program(meson.source_root() / 'subprojects/unity/auto/generate_test_runner.rb')
-else
-gen_test_runner_path = find_program('subprojects/unity/auto/generate_test_runner.rb')
-endif
-
-# Create a generator that we can access from the parent project
-gen_test_runner = generator(gen_test_runner_path, output: '@BASENAME@_Runner.c', arguments: ['@INPUT@', '@OUTPUT@'] )
+# Create a generator that can be used by consumers of our build system to generate
+# test runners.
+gen_test_runner = generator(
+  find_program(meson.project_source_root() / 'auto' / 'generate_test_runner.rb'),
+  output: '@BASENAME@_Runner.c',
+  arguments: ['@INPUT@', '@OUTPUT@']
+)
--- a/meson.build
+++ b/meson.build
@@ -6,6 +6,17 @@
 #
 project('unity', 'c',
   license: 'MIT',
+
+  # Set project version to value extracted from unity.h header
+  version: run_command(
+    [
+      find_program('python', 'python3'),
+      'auto' / 'extract_version.py',
+      'src' / 'unity.h'
+    ],
+    check: true
+  ).stdout().strip(),
+
   # `meson.project_source_root()` introduced in 0.56.0
   meson_version: '>=0.56.0',
   default_options: [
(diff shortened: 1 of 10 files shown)


amcnulty-fermat:
@maintainers, many thanks for the CI approval. Given that it seems to pass the CI check I will mark this as ready for review.

amcnulty-fermat:
I can't set reviewers but it would be good to get feedback from @optorres, and @pmembrey as their Meson build scripts are being modified here. @eli-schwartz, being a Meson developer, is well-placed to offer good feedback.

meson.build
```diff
@@ -5,21 +5,76 @@
...
 project('unity', 'c',
-    license: 'MIT',
-    meson_version: '>=0.37.0',
-    default_options: ['werror=true', 'c_std=c11'])
+  license: 'MIT',
+
+  # Set project version to value extracted from unity.h header
+  version: run_command(
+    [
+      find_program('python', 'python3'),
```
eli-schwartz, line 13: You don't need to find python here, because the extract_version.py script is runnable on its own as a regular program (it's executable and has a shebang). On Windows, Meson will scan the shebang and detect that it needs to be run with python.

meson.build
```diff
@@ -5,21 +5,76 @@
...
+  # `meson.project_source_root()` introduced in 0.56.0
+  meson_version: '>=0.56.0',
+  default_options: [
+    'werror=true',
+    'c_std=c11'
+  ]
+)
+
+build_fixture = get_option('extension_fixture').enabled()
+build_memory = get_option('extension_memory').enabled()
```
eli-schwartz, line 29: You're just using these like booleans, it seems misleading to allow people to specify either `=auto` or `=disabled` and have auto really just mean disabled.

What about defining the options as type boolean, and just checking the option value directly?

meson.build
```diff
@@ -5,21 +5,76 @@
...
+# Generate pkg-config file.
+pkg = import('pkgconfig')
+pkg.generate(unity_lib,
+  description: 'C Unit testing framework.'
+)
+
+# Create a generator that can be used by consumers of our build system to generate
+# test runners.
+gen_test_runner = generator(
+  find_program(meson.project_source_root() / 'auto' / 'generate_test_runner.rb'),
```
eli-schwartz, line 67:
```suggestion
  find_program('auto/generate_test_runner.rb'),
```

find_program() will already search for local files relative to the current context meson.build file so you don't need to prepend an absolute path.

This also means you don't need to bump the minimum version just to use this.

eli-schwartz (changes requested):
My pleasure. :) Some comments which are largely copied from the original PR:

meson.build
amcnulty-fermat: Done. This simplifies things a lot.

amcnulty-fermat:
Thanks @eli-schwartz for your review! 

I have implemented your feedback. In the process of lowering the minimum required version of Meson, I have made some other changes to the toplevel `meson.build` in order to avoid generating warnings. Please let me know what you think.

Given that most users of Meson with Unity will be using it as a subproject, I thought it wise to avoid as many warnings as possible to avoid causing noise or confusion for users. Let me know if this is the right call or not, @mvandervoord.

eli-schwartz:
You need at least 0.47.0 for the check: argument to run_command(), but that version is practically everywhere anyway.

It's not present in distros that are as old as, for example, Debian Jessie or Ubuntu 18.04, but I don't think that's a serious downside.

amcnulty-fermat:
> You need at least 0.47.0 for the check: argument to run_command(), but that version is practically everywhere anyway.

Interesting. Currently when I set `meson_version: '0.37.0'` and specify `check: true` in `run_command()`, Meson 1.0.0 does not seem to generate a warning. Should it generate one?

Using other functionality like `summary()` and `dict` while setting `meson_version` to `'0.37.0'` did cause Meson to warn about there being a mismatch between the features used and the version of mesion required.

eli-schwartz:
I think that just happens because it's parsed before the meson_version is set, since it's inline in the project() function. :) It's a hard problem to solve, but then again, we don't typically recommend inlining functions into project() -- with the exception of, well, the "version" kwarg (run_command() or files())

amcnulty-fermat:
> I think that just happens because it's parsed before the meson_version is set, since it's inline in the project() function. :) It's a hard problem to solve, but then again, we don't typically recommend inlining functions into project() -- with the exception of, well, the "version" kwarg (run_command() or files())

So I was curious about this and modified the `project()` call to make `meson_version` be the first kwarg, thus parsed before the `run_command`, and in this case as well Meson does not issue a warning. It seems as though in [the implementation](https://github.com/mesonbuild/meson/blob/4d7a72b6db181f09034051ef703e151607784ea8/mesonbuild/interpreter/interpreter.py#L747), it should issue one, right?

Thank you for your time and insight!

eli-schwartz:
No, the issue is that run_command() is evaluated before project() is evaluated, and then project() just sees a string. :)

amcnulty-fermat:
> No, the issue is that run_command() is evaluated before project() is evaluated, and then project() just sees a string. :)
> 
> The warning is issued if the run_command occurs on any line after the project() function has run.
> 
> Possible solutions for handling this could include double parsing or remembering non-issued warnings and repeating them later.

Ah that makes sense. Thanks for clarifying it for me.

Could a Unity maintainer take a look at this PR and let me know if there's anything left to do before it's ready for merging?

PR #659, opened 2023-02-14
Parse test execution time in unity fixture output

fy666:
When using unity fixture with time output flag set, `parse_output.rb` did not parse the execution time.

### DIFF
--- a/auto/parse_output.rb
+++ b/auto/parse_output.rb
@@ -131,10 +131,11 @@ class ParseOutput
   def test_passed_unity_fixture(array)
     class_name = array[0]
     test_name  = array[1]
+    test_time = get_test_time(array[array.length - 1])
     test_suite_verify(class_name)
-    printf "%-40s PASS\n", test_name
+    printf "%-40s PASS %10d ms\n", test_name, test_time
 
-    push_xml_output_passed(test_name) if @xml_out
+    push_xml_output_passed(test_name, test_time) if @xml_out
   end
 
   # Test was flagged as having failed so format the output.
--- a/auto/parse_output.rb
+++ b/auto/parse_output.rb
@@ -143,13 +143,14 @@ class ParseOutput
   def test_failed_unity_fixture(array)
     class_name = array[0]
     test_name  = array[1]
+    test_time = get_test_time(array[array.length - 1])
     test_suite_verify(class_name)
     reason_array = array[2].split(':')
     reason = reason_array[-1].lstrip.chomp + ' at line: ' + reason_array[-4]
 
-    printf "%-40s FAILED\n", test_name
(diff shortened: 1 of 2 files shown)


fy666:
Hello @mvandervoord,

I'm very sorry to bother you but I was wondering if I did something wrong or I missed some steps for submitting this PR (as I got no reaction) ?

Best regards,

auto/parse_output.rb
```diff
@@ -131,10 +131,11 @@ def prepare_fixture_line(line)
   def test_passed_unity_fixture(array)
     class_name = array[0]
     test_name  = array[1]
+    test_time = get_test_time(array[array.length - 1])
```
Letme, line 134: I assume this needs to be enabled by something, as by default we do not have the timing information.

fy666: Thank you for reviewing :)
Yes indeed, the timing output has to be enabled using `UNITY_INCLUDE_EXEC_TIME` flag

Letme: so shouldnt there be an if statement, otherwise you are just accessing test_name in the worst case?

Letme:
Why is this printout only for passed and not for other cases (failed)?

Letme:
in line 212 failed case also has time, so I am a bit puzzled why this also does not have it (as well as ignored).

auto/parse_output.rb
fy666: Yes, if the execution time is not present get_test_time will receive a string that wont match the  "( ms)" pattern and return 0 (that strategy was already applied for normal unity output)

fy666:
> in line 212 failed case also has time, so I am a bit puzzled why this also does not have it (as well as ignored).

I updated the PR to parse time for fail and ignore cases (for consistency)

However, please note that  in unity fixture, the execution time is only printed for passing tests: https://github.com/ThrowTheSwitch/Unity/blob/master/extras/fixture/src/unity_fixture.c#LL298C13-L298C34
I can update that as well if you want

Letme:
> However, please note that in unity fixture, the execution time is only printed for passing tests:
> I can update that as well if you want

This I can't decide, but I like uniform behavior among features. Strange that other outputs had time in as well - so maybe that is inconsistency which we need to fix. Either have execution time on passing testcases only, or on all (but everywhere the same).

fy666:
> > However, please note that in unity fixture, the execution time is only printed for passing tests:
> > I can update that as well if you want
> 
> This I can't decide, but I like uniform behavior among features. Strange that other outputs had time in as well - so maybe that is inconsistency which we need to fix. Either have execution time on passing testcases only, or on all (but everywhere the same).

Sorry for the delay.
I just added a commit to make unity fixture output the execution time for failed and ignored tests as well.

Issue #661, opened 2023-02-21
Enabling installation by default breaks projects using unity as meson subproject


nirs:
Since #658, when a project is consuming unity via meson subproject, installing the project
also install unity headers, static library, and pkgconfig file. I think this is always
unwanted since unity is a testing tool and never a runtime requirement.

## A minimal reproducer

```
$ tree
.
├── meson.build
└── subprojects
    └── unity.wrap

$ cat meson.build 
project('unity-repr', 'c')
dependency('unity', fallback: ['unity', 'unity_dep'])

$ cat subprojects/unity.wrap 
[wrap-git]
url = https://github.com/ThrowTheSwitch/Unity.git
revision = head
```

Example build and install:
```
$ meson setup build
The Meson build system
Version: 1.0.0
Source dir: /home/nsoffer/tmp/unity-repr
Build dir: /home/nsoffer/tmp/unity-repr/build
Build type: native build
Project name: unity-repr
Project version: undefined
C compiler for the host machine: ccache cc (gcc 12.2.1 "cc (GCC) 12.2.1 20221121 (Red Hat 12.2.1-4)")
C linker for the host machine: cc ld.bfd 2.38-25
Host machine cpu family: x86_64
Host machine cpu: x86_64
Found pkg-config: /usr/bin/pkg-config (1.8.0)
Found CMake: /usr/bin/cmake (3.25.2)
Run-time dependency unity found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency unity
Cloning into 'unity'...
remote: Enumerating objects: 6463, done.
remote: Counting objects: 100% (686/686), done.
remote: Compressing objects: 100% (340/340), done.
remote: Total 6463 (delta 421), reused 553 (delta 342), pack-reused 5777
Receiving objects: 100% (6463/6463), 6.79 MiB | 4.98 MiB/s, done.
Resolving deltas: 100% (4005/4005), done.

Executing subproject unity 

unity| Project name: unity
unity| Project version: 2.5.4
unity| C compiler for the host machine: ccache cc (gcc 12.2.1 "cc (GCC) 12.2.1 20221121 (Red Hat 12.2.1-4)")
unity| C linker for the host machine: cc ld.bfd 2.38-25
unity| Program auto/generate_test_runner.rb found: YES (/home/nsoffer/tmp/unity-repr/subprojects/unity/auto/generate_test_runner.rb)
unity| Build targets in project: 1
unity| Subproject unity finished.

[...]

PR #662, opened 2023-02-21
Don't install anything when building as subproject

nirs:
When a project is consuming unity as as subproject, unity headers, static library and pkg config files are installed by `meson install`.

This can be fixed by using `meson install --skip-subprojects`, but this must be repeated in all the distros packaging a project.

Fixed by disabling install when building as a subproject.

Fixes: #661


amcnulty-fermat:
My apologies for having caused this issue with my PR. If ever anything I have merged causes difficulties, please ping me and I will work to fix it.

PR #674, opened 2023-05-16
Added support for noreturn functions.

informatimago:
This patch adds macros to support noreturn mocked functions:

Added TEST_PROTECT_NORETURN(), TEST_DO_NOT_RETURN(), and TEST_NOT_RETURNING(call) to test for non-returning functions.
Mock functions that are declared noreturn shall use TEST_DO_NOT_RETURN() instead of return.

More details in the following CMock PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ Makefile
 CMakeFiles
 CMakeCache.txt
 !unityConfig.cmake
+*~
--- a/docs/UnityAssertionsReference.md
+++ b/docs/UnityAssertionsReference.md
@@ -222,6 +222,50 @@ This can be useful for outputting `INFO` messages into the Unity output stream
 without actually ending the test. Like pass and fail messages, it will be output
 with the filename and line number.
 
+#### `TEST_PROTECT_NORETURN()`
+
+This macro prepares an environment to test a function that is declared `noreturn`.
+
+```
+    if(TEST_PROTECT_NORETURN()){
+    }
+```
+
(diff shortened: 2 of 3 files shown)


mvandervoord:
I'm not clear on what this PR is doing that can't be handled already with the existing macros.

Am I understanding correctly that the use-case is that you want to mock a function which is supposed to never return. If the mock is called properly, the test should immediately end with a pass. If we continue past that line (returning back to the main source, for example) it should fail. yes?

```
// Code under test
void run_me(int i) {
  if (i== 0) {
    never_returns();
  }
  
  report_error(i);
}
```

```
// Tests
int never_return_calls_expected= 0;

void never_returns_callback(int num) {
  (void)num; //ignore it. we're not using it.

  if (never_return_calls_expected > 0) {
    never_return_calls_expected--;
    TEST_PASS("callback called");
  }
}

void setUp() {
  never_returns_Stub(never_returns_callback);
  never_return_calls_expected= 0;
}

void tearDown() {
  TEST_ASSERT_EQUAL(0, never_return_calls_expected, "We were expecting calls to never return that didn't happen");
}

void test_run_me_SHOULD_report_error_if_one_is_given(void) {
  report_error_Expect(42);

  run_me(42);
}

void test_run_me_SHOULD_start_our_infinite_loop_if_no_errors(void) {
  never_return_calls_expected=1;
  run_me(0);
}
```

It's completely testable using the current features, I believe. I think this pattern is very useful, though. In fact, I think we should add a CMock plugin for accomplishing this same task?

Am I understanding the goal correctly? Missing anything?

PR #686, opened 2023-07-14
Support shared library build

uilianries:
So far is only possible to build Unity as static library, but actually it would be nice if was possible to not enforce only static, but enabling both configurations.

This pull request removes the STATIC directive from `add_library` and exports all symbols when building on Windows, only when as shared.

Plus, I added an entire workflow to validate Unity with CMake, building as static and shared library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ cmake_minimum_required(VERSION 3.12)
 # Read src/unity.h file and get project version from it
 set(UNITY_HEADER "src/unity.h")
 
-file(STRINGS "${UNITY_HEADER}" UNITY_HEADER_CONTENT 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -144,16 +145,23 @@ target_compile_options(${PROJECT_NAME}
         >
 )
 
+target_compile_definitions(${PROJECT_NAME}
+    PRIVATE
+        # Enable shared library
+        $<$<BOOL:${BUILD_SHARED_LIBS}>:UNITY_SKIP_DEFAULT_RUNNER>
```
AJIOB, line 151: What about adding option for `BUILD_SHARED_LIBS`, as described [in cmake docs](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html)?

Issue #616, opened 2022-06-14
Prevent unity from bailing on failed tests


lmapii:
I'm using unity on target and it works beautifully. However, when I have a test that fails, then all the other tests are skipped without any output due to the `UNITY_FAIL_AND_BAIL` macro in the assertion functions (leading to `TEST_ABORT`).

So far the only way to prevent this is to define `UNITY_EXCLUDE_SETJMP_H` in my `unity_conf.h` file, which might have some impact on mocking. Is there no other way to do this? I've never seen this behaviour when using *ceedling*, there the test execution continues on failures.

mvandervoord:
Hi.

Unity aborts the current test on a failure (when used by itself or with Ceedling). It's the "current test" portion that is important here. In either case, it should move on to the next test.

My suspicion is that something about your structure on your target tests isn't quite right. Maybe all your tests are getting called from a single RUN_TEST() call? Maybe you don't have RUN_TEST() wrappers around your tests at all? Depending on your target, people also can run into situations where their tests run into memory problems, and Unity has no way to recover from it and so execution just stops.

Are you using a generated test runner or are you creating your own? If the latter, it'd be good to see an example of what that looks like. 

Are you using the default RUN_TEST macro, or have you created your own? If the latter, it'd be good to see that as well. :)

lmapii:
Sorry, providing more information: My test runner is created manually, it runs within an OS task:

```c
void Task00(...)
{
    // <snip>
    // UNITY_BEGIN() is called from the startup hooks
    RUN_TEST(TestDopFun);
    RUN_TEST(TestDopOwner);
    RUN_TEST(TestPwmOwner);
    RUN_TEST(TestPwmFun);
    RUN_TEST(TestLpoOwner);
    RUN_TEST(TestLpoFun);
    UNITY_END();
    __asm("debug");
}
```
When everything is OK I'm seeing all results correctly (handled via the debugger interface):
```
../Application/TestOwner/src/TestDop.c:50:TestDopFun:PASS
../Application/TestOwner/src/TestDop.c:51:TestDopOwner:PASS
../Application/TestOwner/src/TestPwm.c:52:TestPwmOwner:PASS
../Application/TestOwner/src/TestPwm.c:53:TestPwmFun:PASS
../Application/TestOwner/src/TestLpo.c:54:TestLpoOwner:PASS
../Application/TestOwner/src/TestLpo.c:55:TestLpoFun:PASS

-----------------------
[...]

PR #695, opened 2023-08-29
Allow user-defined TEST_PROTECT & TEST_ABORT macros

fkjagodzinski:
This update covers use cases where:
- Unity is used to provide the assertion macros only, and an external test harness/runner is used for test orchestration/reporting, e.g. [Mbed OS](https://github.com/ARMmbed/mbed-os/blob/master/features/frameworks/unity/README.md#unity-unit-testing-for-c).
- Calling `longjmp` on a given platform is possible, but has a platform-specific (or implementation-specific) set of prerequisites, e.g. privileged access level. Which has been mentioned in https://github.com/ThrowTheSwitch/Unity/issues/616#issuecomment-1155219717.

Enable project-specific customisation of `TEST_PROTECT` and `TEST_ABORT` macros:
- Use the user-defined `UNITY_TEST_ABORT` if available; fall back to default behaviour otherwise.
- Use the user-defined `UNITY_TEST_PROTECT` if available; fall back to default behaviour otherwise.
- These may be defined independently.

### DIFF
--- a/docs/UnityConfigurationGuide.md
+++ b/docs/UnityConfigurationGuide.md
@@ -399,6 +399,68 @@ _Example:_
 #define UNITY_EXCLUDE_SETJMP
 ```
 
+#### `UNITY_TEST_PROTECT`
+
+#### `UNITY_TEST_ABORT`
+
+Unity handles test failures via `setjmp`/`longjmp` pair by default. As mentioned above, you can disable this with `UNITY_EXCLUDE_SETJMP`. You can also customise what happens on every `TEST_PROTECT` and `TEST_ABORT` call. This can be accomplished by providing user-defined `UNITY_TEST_PROTECT` and `UNITY_TEST_ABORT` macros (and these may be defined independently).
+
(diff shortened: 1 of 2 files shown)


fkjagodzinski:
Marking this as a draft because the patch is missing docs update. :slightly_smiling_face:

Also, please let me know if this path is acceptable, then I'll update the relevant markdown files.

mvandervoord:
Hi. Thanks for the patch! I appreciate that you've matched the pattern of usage we have for other override-able options in the project. If you update the configuration docs with this, I'd say you have a great feature here. :)

fkjagodzinski:
I've updated the docs. I think the macro descriptions are OK, but my view is rather biased, so I'd appreciate any reviews. :wink:

fkjagodzinski:
Hi there. It's been a while so I took another look at the docs, and I'd say both the macro descriptions and the code snippets still make sense. @mvandervoord, could you review the PR when you get a chance?

PR #706, opened 2023-12-02
Add help option to test command line args

Skinner927:
When test binaries are run with unknown options or with the standard -h option, a help menu will print all available options.

This is much more convenient than having to dig through unity.c to find every option.


HaorongX:
Great idea. I also found it confusing when I have to check `unity.c` to find these options. I think the usage should be added to docs at the same time. ^_^

src/unity.c
```diff
@@ -2329,6 +2329,17 @@ int UnityParseOptions(int argc, char** argv)
                     UnityPrint("ERROR: Unknown Option ");
                     UNITY_OUTPUT_CHAR(argv[i][1]);
                     UNITY_PRINT_EOL();
+                    /* fall-through to display help */
+                case 'h':
+                    UnityPrint("Options: "); UNITY_PRINT_EOL();
+                    UnityPrint("-l        List all tests"); UNITY_PRINT_EOL();
+                    UnityPrint("-f TEST   Only run tests with TEST in the name"); UNITY_PRINT_EOL();
```
mvandervoord, line 2336: Maybe change the wording to make it obvious what the function letter is? Something like "Filter to run only tests with TEST in the name"?

src/unity.c
```diff
@@ -2329,6 +2329,17 @@ int UnityParseOptions(int argc, char** argv)
...
+                    UnityPrint("-n TEST   Only run tests with TEST in the name"); UNITY_PRINT_EOL();
```
mvandervoord, line 2337: I'd not include the `n` option... or if I did include it, mark it as `deprecated`?

test/tests/test_generate_test_runner.rb
```diff
@@ -1158,9 +1158,43 @@
       :to_pass => [ ],
       :to_fail => [ ],
       :to_ignore => [ ],
-      :text => [ "ERROR: Unknown Option z" ],
+      :text => [
+        "ERROR: Unknown Option z",
+        "Options:",
+        "-l        List all tests",
+        "-f TEST   Only run tests with TEST in the name",
+        "-n TEST   Only run tests with TEST in the name",
```
mvandervoord, line 1166: see above

mvandervoord (changes requested):
Thanks for the awesome addition! I really appreciate the work. I have a couple minor suggested tweaks, if you don't mind?

src/unity.c
Skinner927: Updated the options. If you'd prefer to choose exactly what to say I won't be offended.

P.S. TY for the fast review.

test/tests/test_generate_test_runner.rb
Skinner927: Marked as depreciated. I can remove it if you'd prefer.

PR #759, opened 2024-11-01
Moved the unity_config.h from examples to the src directory

FrankVanris2:
I've moved the `unity_config.h` file from the `examples` directory to the `src` directory to improve the compilation process on PlatformIO. Keeping it in the `examples` directory was causing unnecessary complications. Placing it in the `src` directory simplifies our workflow and makes things more efficient for everyone.


mvandervoord:
Hi. Thanks for the submission. 

I suspect you're not understanding how Unity is usually used. Your niche use-case where the sample `unity_config.h` that exists in the example directory happens to be useful to you is not common at all. It's an example, nothing more/less. In most cases, `unity_config.h` is not used at all. Of the remaining cases, the great majority of people build their own `unity_config.h` to have the features they specifically need for their environment and toolchains.

I do appreciate that you're looking to fill a need for some users and that you're a person who is willing to share solutions with others. If we can think of a way of helping people like you that doesn't make it inconvenient for the majority of the users, I'd love to consider it. 

At this time, I'm going to have to decline this PR. In it's current form, it helps a small fraction of our users and inserts an obstacle to be worked around for the majority of the users.

PR #762, opened 2025-01-08
New command line options for fixtures

koy-rehme-bae:
:pineapple:

This merge request adds three command-line options to fixtures:

* `-N` -- Like `-n`, but will only select a single test that matches the name exactly.
* `-G` -- Like `-g`, but will only select a single group that matches the name exactly.
* `-d` -- Dry runs the tests. I find this useful to know exactly which tests will be run after applying filters.

If both `-n` and `-N` are provided, selected tests match either one.  The same applies to `-g` and `-G`.

I'm sure there will be some bikeshedding on the characters used for the options; feedback is welcome.

### DIFF
--- a/extras/fixture/src/unity_fixture.c
+++ b/extras/fixture/src/unity_fixture.c
@@ -46,21 +46,25 @@ int UnityMain(int argc, const char* argv[], void (*runAllTests)(void))
     return (int)Unity.TestFailures;
 }
 
-static int selected(const char* filter, const char* name)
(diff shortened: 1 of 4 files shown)


mvandervoord:
This is looking great. I think it's also worthwhile to add the new features to `unity_fixture_Test.c`. You can see an example of testing `-n` in `TEST(UnityCommandOptions, OptionSelectTestByName)`

koy-rehme-bae:
@mvandervoord Tests added as suggested.

Issue #741, opened 2024-07-11
Some codes failed `-Werror=float-equal` compiler flag, please fix it.


jymchng:
Example error message.

```
/root/c_projects/c_template_two/build/coverage/_deps/unity-src/src/unity.c: In function ‘UnityPrintFloat’:
/root/c_projects/c_template_two/build/coverage/_deps/unity-src/src/unity.c:355:16: error: comparing floating-point with ‘==’ or ‘!=’ is unsafe [-Werror=float-equal]
  355 |     if (number == 0.0f)
      |                ^~
/root/c_projects/c_template_two/build/coverage/_deps/unity-src/src/unity.c:422:52: error: comparing floating-point with ‘==’ or ‘!=’ is unsafe [-Werror=float-equal]
  422 |         if ((n & 1) && (((UNITY_DOUBLE)n - number) == 0.5f))
      |                                                    ^~
```

PR #770, opened 2025-03-09
Fix failed build with compile flag: -Werror=float-equal (#741)

artem-sharipov:
This is the fix for issue #741.

Briefly, compilation with the `-Werror=float-equal` flag **failed**.

The cause of the errors was the direct comparison of a floating-point number to another number in three places in the source, two of which are listed in the issue #741. The third error location is the implementation of the number check for NaN - `UNITY_IS_NAN(n)`. In fact, a check of the form `n != n` allows us to determine whether a number `n` is a NaN when there is no way to use `isnan()`. In this case, a direct comparison of a floating-point number is acceptable because the number is being compared to itself, but the compiler flag does not make an exception in this case, so the implementation of the NaN check has been changed.

Elsewhere in the code, the value `epsilon` was introduced to solve the problem. `epsilon` is the error by which one number can be equal to another number. This error occurs because of the finite precision of the computer system's representation of real numbers. For this reason, floating point numbers cannot be directly compared using the `==` and `!=` operators. The numerical `epsilon` corresponds to the precision of the representation of floating point numbers, which is determined by the data type (`float` or `double`). 

The presence of the problem was verified as follows:
1. in the `examples/example_1/makefile` I added the line `CFLAGS += -Werror=float-equal`;
2. ran make in the directory `examples/example_1/`;
3. saw that the build failed with the errors described in the issue #741.

After making the changes to the code, I repeated step 2 and checked that there were no more build errors.

### DIFF
--- a/src/unity.c
+++ b/src/unity.c
@@ -337,10 +337,12 @@ void UnityPrintFloat(const UNITY_DOUBLE input_number)
     static const int sig_digits = 9;
     static const UNITY_INT32 min_scaled = 100000000;
     static const UNITY_INT32 max_scaled = 1000000000;
+    static const UNITY_DOUBLE epsilon = UNITY_DOUBLE_PRECISION;
 #else
     static const int sig_digits = 7;
     static const UNITY_INT32 min_scaled = 1000000;
     static const UNITY_INT32 max_scaled = 10000000;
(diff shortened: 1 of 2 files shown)


artem-sharipov:
The force push only for verified signature of commit.

PR #777, opened 2025-04-18
meson: Add '@EXTRA_ARGS@' to generate_test_runner arguments

dloidolt:
Extended the argument list in meson generator for the test runner strict by adding '@EXTRA_ARGS@'. This change enables the test runner script to accept additional configuration parameters alongside '@INPUT@' and '@OUTPUT@', increasing its flexibility.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -76,5 +76,5 @@ endif
 gen_test_runner = generator(
   find_program('auto/generate_test_runner.rb'),
   output: '@BASENAME@_Runner.c',
-  arguments: ['@INPUT@', '@OUTPUT@']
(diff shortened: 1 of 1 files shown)


Letme:
Hi @dloidolt,
Can you please give me a simple meson.build file, so that I can see how you manage to genrate the runner and do mocks from meson. Because I am abandoning meson because I had to run pre-meson script to generate mocks, so that meson would then get the file list to compile.

dloidolt:
Hi @Letme, 
Here is my meson.build file to generate the test runners: https://github.com/dloidolt/airs-compression/blob/c1b44753fb1dd701bc752a781d6f494c45235dba/test/meson.build#L41
I am currently not using mocks. 
I hope this helps you.

Issue #782, opened 2025-05-28
Inconsistent enum definition for UNITY_COMPARISON_T


rstahn:
My static code analysis tool (PC-Lint) is warning me about the definition of the enum `UNITY_COMPARISON_T`:

```
typedef enum
{
    UNITY_WITHIN           = 0x0,
    UNITY_EQUAL_TO         = 0x1,
    UNITY_GREATER_THAN     = 0x2,
    UNITY_GREATER_OR_EQUAL = 0x2 + UNITY_EQUAL_TO,
    UNITY_SMALLER_THAN     = 0x4,
    UNITY_SMALLER_OR_EQUAL = 0x4 + UNITY_EQUAL_TO,
    UNITY_NOT_EQUAL        = 0x0,
    UNITY_UNKNOWN
} UNITY_COMPARISON_T;
```

Indeed, the value of `UNITY_NOT_EQUAL` explicitly matches `UNITY_WITHIN` and the value of `UNITY_UNKNOWN` implicitly matches `UNITY_EQUAL_TO`. This is currently noncritical since I cannot find any references to neither `UNITY_WITHIN` nor `UNITY_UNKNOWN` outside the definition of `UNITY_COMPARISON_T`. 

Nevertheless the definition looks broken to me. I would like to propose the following changes:
- Remove the unused values `UNITY_WITHIN` and `UNITY_UNKNOWN`
- Change the value of `UNITY_NOT_EQUAL` to 0x8 (because enum values of `UNITY_COMPARISON_T` are supposed to be combinable)

PR #783, opened 2025-05-28
Cleanup definition of UNITY_COMPARISON_T in unity_internals.h

rstahn:
Proposed changes for fixing #782 by removing unused values UNITY_WITHIN and UNITY_UNKNOWN and updating value UNITY_NOT_EQUAL

### DIFF
--- a/src/unity_internals.h
+++ b/src/unity_internals.h
@@ -487,14 +487,12 @@ typedef enum
 
 typedef enum
 {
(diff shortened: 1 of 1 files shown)


rstahn:
Thanks for accepting the PR and the additional feedback. 

I enjoy using Unity & CMock in my daily professional work and I am willing to pay back at least a small dividend 👍

PR #784, opened 2025-06-05
fix: Correct UINT max value handling

burakutkuc:
Related to https://github.com/ThrowTheSwitch/CMock/issues/509

In addition:

The following guard clause:

```
if ((threshold == actual) && (compare & UNITY_EQUAL_TO)) { return; }
if ((threshold == actual))                               { failed = 1; }
```

was replaced with:

```if ((threshold == actual) && !(compare & UNITY_EQUAL_TO)) { failed = 1; }```

The goal was to align the behavior of the `UnityAssertGreaterOrLessOrEqualNumber `function with its name.
In the previous implementation, the function would return early without handling "greater or equal" or "less or equal" cases.
With this update, the function properly allows those comparisons when no failure condition is met.


mvandervoord:
Hi @burakutkuc -- my apologies for taking so long to get back to you on this. I really appreciate that you've chosen to help this project! 

Running the regression tests, it appears there are still some changes to make (see links above). Do you have time to take a look at them?

burakutkuc:
Hi @mvandervoord , no worries at all. thanks for getting back.

 I’ve reviewed the test results and fixed the issue. it should be all good now.

Issue #793, opened 2025-07-18
Generated runners fail to compile when command line arguments are enabled for parameterized tests with strings or newlines


jeronstone:
# Problem

Given TEST_CASE fixtures with string parameters,

```{c}
TEST_CASE("SOMESTRINGHERE")
```
or newlines,
```{c{
TEST_CASE("string.hello.1.2.3.4", REALLY_LONG_ENUM_NAMES2, REALLY_LONG_ENUM_NAMES3,
          NEWLINE_SOME_ENUM)
```

AND `UNITY_USE_COMMAND_LINE_ARGS` is set and compiled with `-DUNITY_USE_COMMAND_LINE_ARGS`, and generate_test_runner.rb is run with `--cmdline_args=1` and `--use_param_tests=1`,

The generated runner `.c` files do not compile.

Example error:
```{c}
  170 |       UnityPrint("  test_cool_name(SOME_ENUM, "12345")");
      |                                         ^~~~~~~~~~~~~~~~~~~~
/path/build/default/test/test_file_runner.c:170:77: error: expected ‘)’ before numeric constant
```

# Probable Cause

The output from [line 472 of generate_test_runner.rb](https://github.com/ThrowTheSwitch/Unity/blob/bddb1366a2e3bf18f0f626fd532a1388032c0e9f/auto/generate_test_runner.rb#L472) writes the TEST_CASE parameters verbatim to the `_runner.c`. When the parameters include strings, the double quotes (") are written to the runner file, and a C compiler will treat them as the end of the String begun by `UnityPrint`, causing all sorts of compilation failures. 

Newlines are also written verbatim, which introduces a few possibilities for compilation error (e.g. newline causes string to not end).

# Proposed Solution

Replace line 472 with:
```{ruby}
output.puts("      UnityPrint(\"  #{test[:test]}(#{args.gsub('"','').gsub("\n",'')})\");")
``` 

which will remove all double-quotes (") and newline characters from the TEST_CASE parameters. Since these are just prints, the impact of not having the double quotes is minimal, and mostly just a visual change.

There may be better ways to solve this issue though, any thoughts?

KaSroka:
Also encountered this bug. Can we proceed with merging the fix?

sullivanmj:
The fix in the pull request worked for me as well.

Letme:
Since fix is merged, can we close this? @mvandervoord

PR #794, opened 2025-07-18
String test case parameter & command-line argument error fix

jeronstone:
Closes issue in #793


sullivanmj:
This works great, thanks for the fix. One suggestion might be to replace `"` with `\"` for formatting purposes.

PR #804, opened 2025-09-16
Add EOL after test message

skocjan:
If TEST_ASSERT_*_MESSAGE() is used and fails, next message is printed on the same line without EOL. This change should improve the readability of output.


skocjan:
Hi @mvandervoord , thanks for explaination. My configuration is a cross compilation to TI ARM platform and to be honest, it must've been some misinterpretation from my side, because now, when I have it properly integrated, I cannot see the issue anymore. Sorry for the noise and PR can be closed

PR #821, opened 2025-12-26
Enable Bazel test filtering in Unity command-line args

tragisch:
:pineapple:

## Summary
Add Bazel test filtering support via TESTBRIDGE_TEST_ONLY when UNITY_USE_COMMAND_LINE_ARGS is enabled, plus core tests and docs.

## Why
Bazel’s --test_filter sets TESTBRIDGE_TEST_ONLY. Unity previously ignored this, so single‑test runs were not possible.

## What Changed
Support TESTBRIDGE_TEST_ONLY in UnityParseOptions
Add Unity core tests for env‑based filtering and CLI override
Document new behavior and note UNITY_GETENV override option
Update UnityChangeLog.md

## Notes
Requires UNITY_USE_COMMAND_LINE_ARGS

### DIFF
--- a/docs/UnityChangeLog.md
+++ b/docs/UnityChangeLog.md
@@ -18,6 +18,7 @@ Prior to 2008, the project was an internal project and not released to the publi
 New Features:
   
   - Add `-n` comand line option as strict matcher again
+  - Support `TESTBRIDGE_TEST_ONLY` for Bazel `--test_filter` when `UNITY_USE_COMMAND_LINE_ARGS` is enabled
 
 Significant Bugfixes:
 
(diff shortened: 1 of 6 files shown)


mvandervoord:
I feel like I'm missing some important context with this PR... like what Bazel or TestBridge are and why a Unity Test developer would need this?

tragisch:
Thanks for asking.

`bazel test //:<test-target>` runs all tests in the C file associated with that target, which is fine in most cases.

Sometimes, though, you want to run just one specific test from that file—for example, when benchmarking or debugging a single failing test among many others. This is especially useful on embedded devices, where test runs can be slow and are often required for safety-critical applications.

`bazel test --test_filter=...` lets you rerun only that individual unit test.

PR #764, opened 2025-01-22
Fix: update meson options to avoid deprecation warnings

dloidolt:
- Remove quotes from boolean values in meson_options.txt
- see https://mesonbuild.com/Release-notes-for-1-1-0.html#coercing-values-in-the-option-function-is-deprecated

PR #826, opened 2026-03-26
meson: allow passing arguments to `generate_test_runner.rb`

lovetheguitar:
This change enables a user to configure the test generation, e.g. enable parametrized tests like so:

```meson
gen_test_runner.process(
      'test_platform.c' , extra_args: ['--use_param_tests=1']
```

This feature is documented [here](https://mesonbuild.com/Reference-manual_returned_generator.html#generatorprocess) in the meson documentation.

Btw. thanks for the great library. :)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -84,5 +84,5 @@ endif
 gen_test_runner = generator(
   find_program('auto/generate_test_runner.rb'),
   output: '@BASENAME@_Runner.c',
-  arguments: ['@INPUT@', '@OUTPUT@']
+  arguments: ['@INPUT@', '@EXTRA_ARGS@', '@OUTPUT@']
 )


Letme:
@mvandervoord I know it does, but fact is there are many open Pull requests there, where you cannot test them and they probably work, but because there aren't any non-authors confirming they need it, they just get stuck and then soon forgotten. So I understand you need to be pinged, and it is much better if you get pinged only when there is approval and work mostly completed.

lovetheguitar:
@Letme Yes I did, it works fine. :) 

@mvandervoord Great, happy to get&give feedback. Will definitely include some output trying to showcase the difference next time. :+1: 

I also verified #764 :heavy_check_mark: :

## Before
```console
  Subprojects
    unity: YES 1 warnings
```

## After
```console
  Subprojects
    unity: YES
```

Short question regarding releases, is there any procedure or schedule how/when something gets released?
Thanks, I appreciate the responsiveness a lot!

mvandervoord:
We are attempting to get to a more regular release cadence. Ceedling has had the largest backlog of needs for the last couple years, and so much of our focus has been on major cleanup/improvements to that. We're just now getting back to a point where we can focus on the foundation projects again. In all cases, our dream is to make supporting these into our fulltime jobs... but at the moment they are sadly free time driven. ;)

PR #844, opened 2026-08-13
Make test runner code compile using '-Werror=unused-parameter'

swaldhoer:


mvandervoord:
Hi @swaldhoer -- whenever possible, I like to use the non-toolchain-specific solution to problems. If instead of using a gcc-specific pragma, you tried doing the `(void)variable_name;` trick for these variables, does that remove the problem on your toolchain? If so, I'd much prefer to use that, because it will fix it elsewhere also. Otherwise, we'll need to wrap this in #ifdefs for checking to see that it's gcc.

swaldhoer:
Yes that works equally well, I'll update the PR later

mvandervoord:
Thanks, you're awesome! :)

